PR #34, opened 2025-02-28
Add `nnsight`-compatible `ModelSplitter`

gsarti:
## Description

This PR introduces a class `ModelSplitter` inheriting from [`nnsight.models.LanguageModel`](https://nnsight.net/documentation/modeling/#nnsight.modeling.language.LanguageModel) that adding `splits` plus some extra utilities (e.g. error messaging for `.generate` with encoder-only models, compatibility checks etc.)

### DIFF
--- a/interpreto/commons/model_wrapping/model_with_split_points.py
+++ b/interpreto/commons/model_wrapping/model_with_split_points.py
@@ -67,12 +67,17 @@ class ActivationSelectionStrategy(Enum):
 
 
 class ModelWithSplitPoints(LanguageModel):
-    """Generalized NNsight.LanguageModel wrapper around encoder-only, decoder-only and encoder-decoder language models.
+    """Code: [:octicons-mark-github-24: model_wrapping/model_with_split_points.py` ](https://github.com/FOR-sight-ai/interpreto/blob/dev/interpreto/commons/model_wrapping/model_with_split_points.py)
+
+    Generalized NNsight.LanguageModel wrapper around encoder-only, decoder-only and encoder-decoder language models.
     Handles splitting model at specified locations and activation extraction.
 
     Inputs can be in the form of:
+
         * One (`str`) or more (`list[str]`) prompts, including batched prompts (`list[list[str]]`).
+
         * One (`list[int] or torch.Tensor`) or more (`list[list[int]] or torch.Tensor`) tokenized prompts.
+
         * Direct model inputs: (`dic[str,Any]`)
 
     Attributes:
@@ -128,8 +133,8 @@ class ModelWithSplitPoints(LanguageModel):
                 raise InitializationError(
                     "Model autoclass not found.\n"
                     "The model class can be omitted if a pre-loaded model is passed to `model_or_repo_id` "
-                    "param.\nIf an HF Hub ID is used, the corresponding autoclass must be specified in `model_type`.\n"
-                    "Example: ModelWithSplitPoints('bert-base-cased', model_type=AutoModelForMaskedLM, ...)"
+                    "param.\nIf an HF Hub ID is used, the corresponding autoclass must be specified in `model_autoclass`.\n"
+                    "Example: ModelWithSplitPoints('bert-base-cased', model_autoclass=AutoModelForMaskedLM, ...)"
                 )
             if isinstance(model_autoclass, str):
                 supported_autoclasses = get_supported_hf_transformer_autoclasses()
@@ -215,7 +220,7 @@ class ModelWithSplitPoints(LanguageModel):
         """Get intermediate activations for all model split points
 
         Args:
-            inputs (str | list[str] | BatchEncoding): Inputs to the model forward pass before or after tokenzation.
+            inputs (str | list[str] | BatchEncoding): Inputs to the model forward pass before or after tokenization.
             select_strategy (str | ActivationSelectionStrategy): Selection strategy for activations.
 
                 Options are:
@@ -226,7 +231,7 @@ class ModelWithSplitPoints(LanguageModel):
             select_indices (list[int] | tuple[int] | None): Specifies indices that should be selected from the
                 activation sequence. Can be combined with `select_strategy` to obtain several behaviors. E.g.
                 `select_strategy="all", select_indices=mask_idxs` can be used to extract only activations corresponding
-                to [MASK] input ids with shape `(batch, len(masx_idxs), d_model)`, or
+                to [MASK] input ids with shape `(batch, len(mask_idxs), d_model)`, or
                 `select_strategy="flatten", select_indices=0` can be used to extract activations for the `[CLS]` token
                 only across all sequences, with shape `(batch, d_model)`. By default, all positions are selected.
 
@@ -255,6 +260,7 @@ class ModelWithSplitPoints(LanguageModel):
                 if idx == len(self.split_points) - 1:
                     getattr(curr_module, module_out_name).stop()
 
+        # TODO: Ideally we would not need to unpack tuples, check if necessary
         for idx, split in enumerate(self.split_points):
             act = activations.value[idx]
             activations_dict[split] = act[0] if isinstance(act, tuple) else act
@@ -272,12 +278,9 @@ class ModelWithSplitPoints(LanguageModel):
                 )
 
             # Apply selection rule
-            if ActivationSelectionStrategy.is_match(select_strategy, ActivationSelectionStrategy.ALL):
-                if select_indices:
-                    activations_dict[layer] = activations_dict[layer][:, select_indices, :]
+            if select_indices:
+                activations_dict[layer] = activations_dict[layer][:, select_indices, :]
             if ActivationSelectionStrategy.is_match(select_strategy, ActivationSelectionStrategy.FLATTEN):
-                if select_indices:
-                    activations_dict[layer] = activations_dict[layer][:, select_indices, :]
                 activations_dict[layer] = activations_dict[layer].flatten(0, 1)
                 if len(activations_dict[layer].shape) != 2:
                     raise RuntimeError(
--- a/interpreto/commons/model_wrapping/splitting_utils.py
+++ b/interpreto/commons/model_wrapping/splitting_utils.py
@@ -96,8 +96,7 @@ def _get_paths_from_module(model_paths: list[str] | None = None, module: nn.Modu
         raise ValueError("Either a module or a precomputed list of valid model paths should be provided for sorting.")
     if model_paths is None:
         model_paths = []
-        if module is not None:
-            model_paths = list(walk_modules(module))
+        model_paths = list(walk_modules(module))  # type: ignore
     return model_paths
 
 
--- a/interpreto/concepts/base.py
+++ b/interpreto/concepts/base.py
@@ -36,20 +36,19 @@ from typing import Any, Generic, TypeVar
 
 import torch
 from overcomplete.base import BaseDictionaryLearning
-from torch import nn
 
 from interpreto.attributions.base import AttributionExplainer
 from interpreto.commons.model_wrapping.model_with_split_points import ModelWithSplitPoints
 from interpreto.typing import ConceptsActivations, LatentActivations, ModelInput
 
-Module = TypeVar("Module", bound=nn.Module)
-T = TypeVar("T")
+ConceptModel = TypeVar("ConceptModel")
+MethodOutput = TypeVar("MethodOutput")
 
 
 # Decorator that checks if the concept model is fitted before calling the method
-def check_fitted(func: Callable[..., T]) -> Callable[..., T]:
+def check_fitted(func: Callable[..., MethodOutput]) -> Callable[..., MethodOutput]:
     @wraps(func)
-    def wrapper(self: AbstractConceptExplainer, *args, **kwargs) -> T:
+    def wrapper(self: AbstractConceptExplainer, *args, **kwargs) -> MethodOutput:
         if not self.is_fitted or self.split_point is None:
             raise RuntimeError("Concept encoder is not fitted yet. Use the .fit() method to fit the explainer.")
         return func(self, *args, **kwargs)
@@ -57,8 +56,10 @@ def check_fitted(func: Callable[..., T]) -> Callable[..., T]:
     return wrapper
 
 
-class AbstractConceptExplainer(ABC, Generic[Module]):
-    """Abstract class defining an interface for concept explanation.
+class AbstractConceptExplainer(ABC, Generic[ConceptModel]):
+    """Code: [:octicons-mark-github-24: `concepts/base.py` ](https://github.com/FOR-sight-ai/interpreto/blob/dev/interpreto/concepts/base.py)
+
+    Abstract class defining an interface for concept explanation.
     Child classes should implement the `fit` and `encode_activations` methods, and only assume the presence of an
         encoding step using the `concept_model` to convert activations to latent concepts.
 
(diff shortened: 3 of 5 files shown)


interpreto/commons/model_wrapping/model_with_split_points.py
```diff
@@ -0,0 +1,287 @@
...
+                If not specified, it will be instantiated with the default configuration for the model.
+            tokenizer (PreTrainedTokenizer): Custom tokenizer for the loaded model.
+                If not specified, it will be instantiated with the default tokenizer for the model.
+        """
+        if isinstance(model_or_repo_id, str):  # Repository ID
+            if model_autoclass is None:
+                raise InitializationError(
+                    "Model autoclass not found.\n"
+                    "The model class can be omitted if a pre-loaded model is passed to `model_or_repo_id` "
+                    "param.\nIf an HF Hub ID is used, the corresponding autoclass must be specified in `model_type`.\n"
```
AntoninPoche: The error message does not seem up to date. In two spots, you talk about `model_type`, but you seem to have replaced it with `model_autoclass`.

docs/api/concepts/base.md
```diff
@@ -0,0 +1,32 @@
+---
+icon: material/middleware-outline
+---
+
+# Base Classes
+
+::: interpreto.concepts.AbstractConceptExplainer
```
AntoninPoche, line 7: I checked the documentation, and it is super clean. I like it!

I only have one question: Is there a way to add a link to open the source file code from Git Hub directly? I know it shows the code for methods. But when I explore a library, I like to open the source file directly.

interpreto/commons/model_wrapping/splitting_utils.py
```diff
@@ -0,0 +1,126 @@
...
+    return model_paths.index(split)
+
+
+def _get_paths_from_module(model_paths: list[str] | None = None, module: nn.Module | None = None) -> list[str]:
+    if model_paths is None and module is None:
+        raise ValueError("Either a module or a precomputed list of valid model paths should be provided for sorting.")
+    if model_paths is None:
+        model_paths = []
+        if module is not None:
+            model_paths = list(walk_modules(module))
```
AntoninPoche: Unnecessary if, `module is not None` is always `True` after previous lines.

You can replace these 3 lines by: `model_paths = list(walk_modules(module))`

interpreto/commons/model_wrapping/model_with_split_points.py
```diff
@@ -0,0 +1,287 @@
...
+        self,
+        inputs: str | list[str] | BatchEncoding,
+        select_strategy: str | ActivationSelectionStrategy = ActivationSelectionStrategy.ALL,
+        select_indices: int | list[int] | tuple[int] | None = None,
+        **kwargs,
+    ) -> dict[str, LatentActivations]:
+        """Get intermediate activations for all model split points
+
+        Args:
+            inputs (str | list[str] | BatchEncoding): Inputs to the model forward pass before or after tokenzation.
```
AntoninPoche: typo: `tokenzation` -> `tokenization`

interpreto/commons/model_wrapping/model_with_split_points.py
```diff
@@ -0,0 +1,287 @@
...
+
+                Options are:
+
+                * `all`: All sequence activations are returned, keeping the original shape `(batch, seq_len, d_model)`.
+                * `flatten`: Every token activation is treated as a separate element - `(batch x seq_len, d_model)`.
+
+            select_indices (list[int] | tuple[int] | None): Specifies indices that should be selected from the
+                activation sequence. Can be combined with `select_strategy` to obtain several behaviors. E.g.
+                `select_strategy="all", select_indices=mask_idxs` can be used to extract only activations corresponding
+                to [MASK] input ids with shape `(batch, len(masx_idxs), d_model)`, or
```
AntoninPoche: typo: `len(masx_idxs)` -> `len(mask_idxs)`

interpreto/commons/model_wrapping/model_with_split_points.py
```diff
@@ -0,0 +1,287 @@
...
+                # Handle case in which module has .output attribute, and .nns_output gets overridden instead
+                module_out_name = "nns_output" if hasattr(curr_module, "nns_output") else "output"
+                activations.append(getattr(curr_module, module_out_name))
+                # Early stopping at the last splitting layer
+                if idx == len(self.split_points) - 1:
+                    getattr(curr_module, module_out_name).stop()
+
+        for idx, split in enumerate(self.split_points):
+            act = activations.value[idx]
+            activations_dict[split] = act[0] if isinstance(act, tuple) else act
```
AntoninPoche, line 266: It seems that in the case of activations that are tuples, more information is forwarded from just the activation tensors. In this case, isn't the remaining information necessary to forward from these activations?

I know that the nnsight team has yet to answer us on how to forward from activations. However, I wanted to know which part of the code might have to be adjusted.

interpreto/commons/model_wrapping/model_with_split_points.py
```diff
@@ -0,0 +1,287 @@
...
+                    f"(batch_size, sequence_length, model_dim), got {activations_dict[layer].shape}"
+                )
+
+            # Apply selection rule
+            if ActivationSelectionStrategy.is_match(select_strategy, ActivationSelectionStrategy.ALL):
+                if select_indices:
+                    activations_dict[layer] = activations_dict[layer][:, select_indices, :]
+            if ActivationSelectionStrategy.is_match(select_strategy, ActivationSelectionStrategy.FLATTEN):
+                if select_indices:
+                    activations_dict[layer] = activations_dict[layer][:, select_indices, :]
```
AntoninPoche: You do the same operation to select indices with both activation strategies. Therefore, it can be factorized outside of activation strategy cases.

Then if you factorize it, there is nothing left in the `ALL` case. Thus, the code becomes:

````python
            # Apply selection rule
            if select_indices:
                activations_dict[layer] = activations_dict[layer][:, select_indices, :]
            if ActivationSelectionStrategy.is_match(select_strategy, ActivationSelectionStrategy.FLATTEN):
                activations_dict[layer] = activations_dict[layer].flatten(0, 1)
                if len(activations_dict[layer].shape) != 2:
                    raise RuntimeError(
                        f"Invalid output for layer '{layer}'. Expected a torch.Tensor activation with shape"
                        f"(batch_size x sequence_length, model_dim), got {activations[layer].shape}"
                    )
```

interpreto/concepts/__init__.py
AntoninPoche, line 1: @fanny-jourdan argued that having "Overcomplete" in the name is unnecessary and misleading. Indeed, we will make clear in the documentation that it is based on Thomas' library.

More generally, I think our naming on the concept part is a disaster ^^

I do not expect you to find any new names now, but maybe we can place a TODO for better names?

interpreto/concepts/base.py
```diff
@@ -23,245 +23,357 @@
...
+from overcomplete.base import BaseDictionaryLearning
+from torch import nn
 
 from interpreto.attributions.base import AttributionExplainer
-from interpreto.commons.model_wrapping.model_splitter import ModelSplitterPlaceholder
+from interpreto.commons.model_wrapping.model_with_split_points import ModelWithSplitPoints
 from interpreto.typing import ConceptsActivations, LatentActivations, ModelInput
 
+Module = TypeVar("Module", bound=nn.Module)
+T = TypeVar("T")
```
AntoninPoche: I do not understand why you used `Module` instead of `nn.Module`.

What is T?

interpreto/concepts/base.py
```diff
@@ -23,245 +23,357 @@
...
+                f"Got: '{type(activations)}'"
+            )
+        activations_split_points = list(activations.keys())
+        if self.split_point not in activations_split_points:
+            raise ValueError(
+                f"Fitted split point '{self.split_point}' not found in activations.\n"
+                f"Available split_points: {', '.join(activations_split_points)}."
+            )
+
+    def get_and_verify_split_point(self, split_point: str | None) -> str:
```
AntoninPoche: I suggest that the `split_point` becomes a property and this function a setter.

Or another way to ensure it is not modified on the fly (as it would not make sense). Maybe a private attribute with a getter.

interpreto/concepts/base.py
```diff
@@ -23,245 +23,357 @@
...
-        splitted_model (ModelSplitterPlaceholder): The model to apply the explanation on. It should be splitted between at least two parts.
-        split (str): The split in the model where the concepts are encoded from.
-        concept_encoder_decoder (ConceptEncoderDecoder): Concept encoder-decoder
-        is_fitted (bool): Whether the model has been fitted
-        has_differentiable_concept_encoder (bool): Whether the concept encoder is differentiable.
-        has_differentiable_concept_decoder (bool): Whether the concept decoder is differentiable.
+        model_with_split_points (ModelWithSplitPoints): The model to apply the explanation on.
+            It should have at least one split point on which `concept_model` can be fitted.
+        split_point (str): The split point used to train the `concept_model`.
+        concept_model ([BaseDictionaryLearning](https://github.com/KempnerInstitute/overcomplete/blob/24568ba5736cbefca4b78a12246d92a1be04a1f4/overcomplete/base.py#L10)): The model used to extract concepts from the
```
AntoninPoche, line 295: There are `ConceptBottleneckExplainers` which will not have a `concept_model` inheriting from `overcomplete.BaseDictionaryLearning`.

Furthermore, for now, all `overcomplete.BaseDictionaryLearning` are not `torch.nn.Module`, I do not know if it is relevant, but it should be kept in mind.

interpreto/concepts/methods/cockatiel.py
```diff
@@ -28,45 +28,68 @@
...
-        splitted_model (ModelSplitterPlaceholder): Model splitter
-        concept_encoder_decoder (oc_opt.NMF): Overcomplete NMF concept encoder decoder
-        is_fitted (bool): Whether the model has been fitted
-        has_differentiable_concept_encoder (bool): Whether the concept encoder is differentiable.
-        has_differentiable_concept_decoder (bool): Whether the concept decoder is differentiable.
+        model_with_split_points (ModelWithSplitPoints): The model to apply the explanation on.
+            It should have at least one split point on which `concept_model` can be fitted.
+        split_point (str | None): The split point used to train the `concept_model`. Default: `None`, set only when
+            the concept explainer is fitted.
+        concept_model (oc_sae.SAE): An [Overcomplete NMF](https://github.com/KempnerInstitute/overcomplete/blob/main/overcomplete/optimization/nmf.py) encoder-decoder.
```
AntoninPoche: Wrong `concept_model` class in docstring: `oc_sae.SAE` -> `oc_opt.SemiNMF`

interpreto/concepts/methods/overcomplete.py
```diff
@@ -0,0 +1,429 @@
...
+                See the Overcomplete documentation of the provided `concept_model_class` for more details.
+        """
+        if not issubclass(concept_model_class, oc_sae.SAE):
+            raise ValueError(
+                "ConceptEncoderDecoder must be a subclass of `overcomplete.sae.SAE`.\n"
+                "Use `interpreto.concepts.methods.OvercompleteSAEClasses` to get the list of available SAE methods."
+            )
+        self.model_with_split_points = model_with_split_points
+        split: str = self.get_and_verify_split_point(split_point)
+        input_shape = self.model_with_split_points.get_activations(self.model_with_split_points._example_input, split)[
```
AntoninPoche: Here, you pass a `split` to the `get_activations` as the third positional argument after `self` and `inputs`. But the third positional argument of `get_activations` is `select_strategy`.

In fact, the `split` you provided here does not seem to be used anywhere. It might have been undetected if tests were done with a `ModelWithSplitPoints` with only one `split_point`.

interpreto/concepts/methods/overcomplete.py
```diff
@@ -0,0 +1,429 @@
...
+        self.has_differentiable_concept_encoder = False if "NMF" in concept_model_class.__name__ else True
+        self.has_differentiable_concept_decoder = True
+
+    def fit(self, activations: LatentActivations | dict[str, LatentActivations], *, overwrite: bool = False, **kwargs):
+        """Fit an Overcomplete OptimDictionaryLearning model on the given activations.
+
+        Args:
+            activations (torch.Tensor | dict[str, torch.Tensor]): The activations used for fitting the `concept_model`.
+                If a dictionary is provided, the activation corresponding to `split_point` will be used.
+            **kwargs (dict): Additional keyword arguments to pass to the `concept_model`.
```
AntoninPoche, line 440: The `overwrite` argument is missing from the docstring.

docs/api/concepts/base.md
gsarti: According to this: https://stackoverflow.com/questions/75785663/mkdocs-mkdocstrings-add-link-back-to-github-source-code there is no direct support for this functionality, and the proposed third-party lib is archived. I'd suggest we stick to the embedded code blocks overall, but I can add links to the corresponding files on top of each `.md` to facilitate further inspections!

AntoninPoche: A link at the top would be enough I think.

interpreto/commons/model_wrapping/splitting_utils.py
gsarti: It wasn't picked up by the typechecker, fixed and added a `# type: ignore`.

interpreto/commons/model_wrapping/model_with_split_points.py
gsarti: My understanding is that this is an issue derived from the `InterventionProxy` not being turned correctly into outputs when exiting the `trace` context. In all cases I saw, the tuple had a single element, which was the activation tensor. TBD, but it's likely this will change.

interpreto/concepts/__init__.py
gsarti: Agreed! I was also thinking of removing the `Overcomplete` from class names 🙂 glad to hear you agree! I find the `OvercompleteDictionaryLearning` to be the worst tbh, since it's too generic (maybe include `Optim` somewhere to remain consistent?). I'll add the TODOs!

interpreto/concepts/base.py
gsarti: `T` is a generic variable type. The annotation for `check_fitted` expresses that the method does not simply take/return a callable returning `Any`, but explicitly binds the return of the input function to the return of the wrapper (i.e., whatever `T` the function returns, the wrapper will also return).

Regarding the `Module`, the idea is that we are saying `Module` is any class belonging to the `nn.Module` category. Then, we define the abstract class as having a `Generic[Module]` associated with its initialization, corresponding to the type that will be assigned to the `concept_model` attribute. Then, when we define the child class `ConceptBottleneckExplainer`, specifying the inheritance as `AbstractConceptExplainer[BaseDictionaryLearning]` tells the type checker that the `Module` is going to be a `BaseDictionaryLearning` for this class so that it does not complain when we call e.g. `encode` or `decode` from the `concept_model`!

interpreto/concepts/methods/overcomplete.py
gsarti: Fixed, also added `activation_select_strategy` and `activation_select_indices` as init params matching the defaults of `get_activations` to enable customization upon class initialization for now.

interpreto/concepts/methods/cockatiel.py
gsarti: Replaced with `overcomplete.optimization.SemiNMF`. Just a quick note in relation to this: we should **always** use full names/paths like this instead of aliases for the docstrings, including for custom types we define in typing.py. This is to make immediately evident to the user what they are dealing with - we can still use the aliases we create for typing!

interpreto/concepts/base.py
gsarti: This is a good idea! 😄 I think a private attribute is probably the right way to go, will modify!

interpreto/concepts/base.py
gsarti: Ah now I understand also the question above related to the `nn.Module` typevar!

For the first point, I think it's fair to enforce the (very high-level) structure of `BaseDictionaryLearning` for future methods, regardless of whether they are in Overcomplete - it should be generic enough to help ensure a bit of uniformity.

For the second point, I will remove the binding to `nn.Module` for the `Module` typevar class to reflect there can be child classes (like NMF) that are not `nn.Module`.

interpreto/concepts/base.py
gsarti: Changed to reflect that not all `Module` are `nn.Module` children

Issue #33, opened 2025-02-27
Create first concept-bottleneck method


AntoninPoche:
No need for nnsight yet

It should use OverComplete

Some general tests should also be included

PR #38, opened 2025-03-04
Concept Bottleneck Explainers and Overcomplete

AntoninPoche:
## Description

- Introduce the `interpreto.concepts.base.ConceptBottleneckExplainer` (CBE) and update the `AbstractConceptExplainer`.
- Introduce some requirements for the model splitting.
- Inherit from it to wrap methods from [https://github.com/KempnerInstitute/Overcomplete](OverComplete) with `OvercompleteDictionaryLearning` and `OvercompleteSAE`.
- Implementation of Cockatiel as an `OvercompleteDictionaryLearning` with `NMF` as the hardcoded method.
- Tests for CBEs and overcomplete wrappers.

## Do not include

- Attributions for input to concepts or for concepts to output.
- Interpretation methods.

## Did not test

- The use of AMP for SAEs with Overcomplete.
- The Cockatiel class as tests are done for `OvercompleteDictionaryLearning` with NMF.

## Related Issue

#33

## Type of Change

## Checklist

### DIFF
--- a/interpreto/concepts/base.py
+++ b/interpreto/concepts/base.py
@@ -44,11 +44,11 @@ class AbstractConceptExplainer(ABC):
     Attributes:
         splitted_model (ModelSplitterPlaceholder): The model to apply the explanation on. It should be splitted between at least two parts.
         split (str): The split in the model where the concepts are encoded from.
-        fitted (bool): Whether the model has been fitted
-        _differentiable_concept_encoder (bool): Whether the concept encoder is differentiable.
+        is_fitted (bool): Whether the model has been fitted
+        has_differentiable_concept_encoder (bool): Whether the concept encoder is differentiable.
     """
 
-    _differentiable_concept_encoder = False
+    has_differentiable_concept_encoder = False
 
     def __init__(self, splitted_model: ModelSplitterPlaceholder):
         """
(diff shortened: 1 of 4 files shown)


interpreto/concepts/base.py
```diff
@@ -1,60 +1,267 @@
...
-        self.model = model
-        self.concept_extraction_method = concept_extraction_method
+        Verify that the given activations are valid for the concept explainer.
+        That is if the split corresponds to the model splits.
+
+        Args:
+            activations (dict[LatentActivations]): the activations to verify.
+            split: (str | None): The dataset split to use for training the concept encoder. If None, the model is assumed to be a single-split model. And split is inferred from the keys of the activations dict.
+        """
+        if hasattr(self, "split") and self.split is not None:
```
thomas-mullor: If you have a "split" attribute, you overwrite the parameter ?

thomas-mullor: You can use getattr(self, "split", default_value)

interpreto/concepts/base.py
```diff
@@ -1,60 +1,267 @@
...
+            split = self.split
+
+        if isinstance(activations, dict):
+            if len(activations) != len(self.splitted_model.splits):
+                raise ValueError(
```
thomas-mullor: Shouldn't we accept the activations to match only a part of the model splits ?

interpreto/concepts/base.py
```diff
@@ -1,60 +1,267 @@
...
+                    f"Activations should be a dict with {self.splitted_model.splits} keys but got {activations.keys()}."
+                )
+            if split is None:
+                if len(activations) != 1 and len(self.splitted_model.splits) != 1:
+                    raise ValueError("Cannot infer split if the model is a not a single-split model.")
+                split = list(activations.keys())[0]
+        elif split is None:
+            if len(self.splitted_model.splits) != 1:
+                raise ValueError("Cannot infer split if the model is a not a single-split model.")
```
thomas-mullor, line 90: There is a code and a condition duplication, maybe you could consider reorganizing your if/else clauses

interpreto/concepts/base.py
AntoninPoche: I could also raise an error, but if the CBE has a split, then it should not be overwritten by the parameter. Or should it?

interpreto/concepts/base.py
AntoninPoche: Yeah, I am not satisfied either, but this is the best I found. I will try more.

interpreto/concepts/base.py
```diff
@@ -1,60 +1,267 @@
...
 
 
 class AbstractConceptExplainer(ABC):
     """
     Abstract class defining an interface for concept explanation.
+
+    Attributes:
+        splitted_model (ModelSplitterPlaceholder): The model to apply the explanation on. It should be splitted between at least two parts.
+        split (str): The split in the model where the concepts are encoded from.
+        fitted (bool): Whether the model has been fitted
```
gsarti: I'd use `is_fitted` to be even more explicit

interpreto/concepts/base.py
```diff
@@ -1,60 +1,267 @@
...
+        _differentiable_concept_encoder (bool): Whether the concept encoder is differentiable.
```
gsarti: Same here, `is_differentiable_...`. What is a concept encoder exactly? Wouldn't `concept_explainer` make more sense here? Also, for consistency we should probably keep it without leading underscore.

interpreto/concepts/base.py
```diff
@@ -1,60 +1,267 @@
...
+        That is if the split corresponds to the model splits.
+
+        Args:
+            activations (dict[LatentActivations]): the activations to verify.
+            split: (str | None): The dataset split to use for training the concept encoder. If None, the model is assumed to be a single-split model. And split is inferred from the keys of the activations dict.
+        """
+        if hasattr(self, "split") and self.split is not None:
+            split = self.split
+
+        if isinstance(activations, dict):
```
gsarti: Ok for now, but I'd like to have activations as a class that encapsulates serialization methods (which would also make checks easier here)

interpreto/concepts/base.py
AntoninPoche: What I call concept encoder here is the encoder part of the concept encoder decoder. the concept encoder decoder takes in latent activations produce concept activations through its encoder then decodes then into reconstructed latent activation.

This is what we called the $t$ and $t^{-1}$.

The difference with the `ConceptBottleneckExplainer` is that these encoder and decoder do not care about the initial model and just treat the activations.

The concept encoder decoder is mainly a reference to the overcomplete objects.

interpreto/concepts/base.py
AntoninPoche: I agree, dictionaries acted as a placeholder for me.

interpreto/concepts/base.py
AntoninPoche: Therefore, it would be more `has_differentiable...` than `is_differentiable...`.

PR #69, opened 2025-06-27
Llm labels

clayecharlotte:
## Description

New concept interpretation method : automatic labeling with LLM.
- Refactored some common code from `topk_input` into `base` to enable re-use in `llm_labels`.
- Replaced `source` by `use_vocab` : the source is automatically inferred from the input.
- Several sampling strategies for selecting examples to give to the LLM : top, random and quantile-based
- The user can provide a user prompt. A default prompt is available if none is provided.

Limitations :
- For now there is only an OpenAI interface
- No few-shot examples to guide the LLM

Example of explanations :
image

## Type of Change

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "interpreto"
-version = "0.3.1"
+version = "0.3.2"
 description = "Interpretability toolbox for LLMs"
 readme = "README.md"
 requires-python = ">=3.10"


concept_examples.ipynb
AntoninPoche, line 1: The notebook `concept_examples.ipynb` was moved and updated in `docs/notebook/concept_examples.ipynb`.

I suggest you add your cells at the end of the other notebook. This way we can compare the different interpretations.

interpreto/concepts/interpretations/base.py
```diff
@@ -90,15 +93,115 @@ def interpret(
...
+                latent_activations = latent_activations.to(self.concept_model.device)  # type: ignore
+            concepts_activations = self.concept_model.encode(latent_activations)
+            if isinstance(concepts_activations, tuple):
+                concepts_activations = concepts_activations[1]  # temporary fix, issue #65
+            return concepts_activations
+
+        if inputs is not None:
+            activations_dict: dict[str, LatentActivations] = self.model_with_split_points.get_activations(
+                inputs,
+                activation_granularity=self.activation_granularity,  # TODO
```
AntoninPoche: There is an undefined TODO.

Furthermore, the type checker is not happy with some elements in this. In particular, in this line, because `self.activation_granularity` is not defined.

interpreto/concepts/interpretations/base.py
```diff
@@ -90,15 +93,115 @@ def interpret(
...
+            )
+            latent_activations = self.model_with_split_points.get_split_activations(
+                activations_dict, split_point=self.split_point
+            )
+            if hasattr(self.concept_model, "device"):
+                latent_activations = latent_activations.to(self.concept_model.device)  # type: ignore
+            concepts_activations = self.concept_model.encode(latent_activations)
+            if isinstance(concepts_activations, tuple):
+                concepts_activations = concepts_activations[1]  # temporary fix, issue #65
+            return concepts_activations
```
AntoninPoche: Duplicated code.

interpreto/concepts/interpretations/llm_labels.py
```diff
@@ -0,0 +1,505 @@
...
+from jaxtyping import Float
+
+from interpreto import Granularity, ModelWithSplitPoints
+from interpreto.concepts.interpretations.base import BaseConceptInterpretationMethod, verify_concepts_indices
+from interpreto.model_wrapping.llm_interface import LLMInterface, Role
+from interpreto.model_wrapping.model_with_split_points import ActivationGranularity
+from interpreto.typing import ConceptModelProtocol, ConceptsActivations, LatentActivations
+
+
+class SAMPLING_METHOD(Enum):
```
AntoninPoche: We use the CamelCase notations for classes and enums:
`SAMPLING_METHOD` -> `SamplingMethod`

interpreto/concepts/interpretations/llm_labels.py
```diff
@@ -0,0 +1,505 @@
...
+
+    def __init__(
+        self,
+        *,
+        model_with_split_points: ModelWithSplitPoints,
+        concept_model: ConceptModelProtocol,
+        split_point: str | None = None,
+        activation_granularity: ActivationGranularity = ActivationGranularity.TOKEN,
+        llm_interface: LLMInterface,
+        sampling_method: SAMPLING_METHOD,
```
AntoninPoche: Would it make sense to have a default value?

interpreto/concepts/interpretations/llm_labels.py
```diff
@@ -0,0 +1,505 @@
...
+                latent_activations=latent_activations,
+                concepts_activations=concepts_activations,
+            )
+
+        concepts_indices = verify_concepts_indices(
+            concepts_activations=sure_concepts_activations,
+            concepts_indices=concepts_indices,
+        )
+
+        granular_inputs, granular_sample_ids = self._get_granular_inputs(sure_inputs)
```
AntoninPoche: These seem to be the same lines as `TopKInputs`. Would it make sense to factorize it?

I saw that the `self.get_granular_inputs` is not the same, but it seems that the `TopKInputs` one is included in the `LLMLabels` one.

interpreto/concepts/interpretations/llm_labels.py
```diff
@@ -0,0 +1,505 @@
...
+                k_quantile=self.k_quantile,
+            )
+        elif self.sampling_method == SAMPLING_METHOD.RANDOM:
+            inputs_idx = _sample_random(
+                concept_activations=concept_activations,
+                k_examples=self.k_examples,
+            )
+        else:
+            raise NotImplementedError(f"Sampling method {self.sampling_method} is not implemented.")
+        return inputs_idx
```
AntoninPoche: The three submethods do not depend on the class apart from the `k_examples`, and `k_quantiles` arguments.

I suggest you move the `sample_example` method to the `SampleMethod` enum, and likewise for the corresponding three submethods.

This should not be too complicated. It is not mandatory, but it would make sense that sampling is managed by the `SamplingMethod` object.

In the `interpret` method, it would imply changing:
```python
self._sample_examples(
                concept_activations=sure_concepts_activations[:, concept_idx],
            )
```
to
````python
SampleMethod.sample_examples(
    sampling_method=self.sampling_method,
    concept_activations=sure_concepts_activations[:, concept_idx],
    k_examples=self.k_examples,
    k_quantile=self.k_quantile,
```

interpreto/concepts/interpretations/llm_labels.py
clayecharlotte: I can factorize the `self.get_granular_inputs` as it does not impact much the current child classes and it seems quite common, but I think that we should be careful not to over-factorize things !

AntoninPoche: Sure, it's your call.

interpreto/concepts/interpretations/llm_labels.py
clayecharlotte: Another option would be to define sampling methods as classes and ask the user to instantiate them, similarly to LLMInterface. It's cleaner because it avoids many parameters specific to sampling method (ex: `k_quantile`) and it avoids the big if/else method. On the other hand, it may be more complex for the user. What do you think ?

```
sampling_method = TopSampling(k_examples=50)
```
or
```
sampling_method = QuantileSampling(k_examples=50, k_quantiles=5)
```
etc

And then :

```
llm_labeling = LLMLabels(
        ...,   #other params
        llm_interface=llm_interface,
        sampling_method=sampling_method,  #contains k_examples, k_quantiles, ...
)
```
And in `LLMLabels.interpret`:

```
example_idx = self.sampling_method(concept_activations=sure_concepts_activations[:, concept_idx])
```

AntoninPoche: I prefer the previous version because the documentation would have been much easier to understand in the first place. Thus, if the sampling method is something that the user has to instantiate, most users will not use it. It is much easier to explore when it is a parameter in an enum.

The `LLMInterface` is something that the user has no choice but to manipulate. But the sampling method could be left as default, and it might work just fine for the user.

AntoninPoche (approved):

PR #72, opened 2025-07-25
Introduce concept-to-output gradients

AntoninPoche:
## Description

TODO

## Type of Change

## Checklist


interpreto/commons/granularity.py
```diff
@@ -78,11 +78,63 @@ class GranularityAggregationStrategy(Enum):
...
+            case GranularityAggregationStrategy.MEAN:
+                return x.mean(dim=dim, keepdim=True)
+            case GranularityAggregationStrategy.MAX:
+                return x.max(dim=dim, keepdim=True)[0]
+            case GranularityAggregationStrategy.MIN:
+                return x.min(dim=dim, keepdim=True)[0]
+            case GranularityAggregationStrategy.SIGNED_MAX:
+                return x.gather(dim, x.abs().max(dim=dim)[1].unsqueeze(dim))
+            case _:
+                raise NotImplementedError(f"Aggregation strategy {strategy} not implemented.")
```
fanny-jourdan: I don't understand this change. The previous version worked fine, didn't it? And why do you use [0] instead of .values for min and max?

interpreto/commons/granularity.py
```diff
@@ -487,8 +539,8 @@ def aggregate_score_for_gradient_method(
                         aggregated_contribution[:, [aggregation_index]] = tokens_contribution
                     else:
                         # aggregate token contribution for each word/sentence
-                        aggregated_contribution[:, [aggregation_index]] = granularity_aggregation_strategy(
-                            tokens_contribution, dim=1
+                        aggregated_contribution[:, [aggregation_index]] = GranularityAggregationStrategy.aggregate(
+                            tokens_contribution, strategy=granularity_aggregation_strategy, dim=1
```
fanny-jourdan, line 543: You change it here by adding .aggregate, but you don't change it in the functions where it is used to aggregate gradient-based attribution methods?

interpreto/concepts/base.py
```diff
@@ -322,7 +331,6 @@ def __init__(
         """
         self.concept_model: BaseDictionaryLearning
         super().__init__(model_with_split_points, concept_model, split_point)
-        self.has_differentiable_concept_decoder = False
```
fanny-jourdan, line 325: Why change the format of the has_differentiable_concept_decoder argument?

interpreto/concepts/interpretations/topk_inputs.py
```diff
@@ -68,7 +68,7 @@ class TopKInputs(BaseConceptInterpretationMethod):
         k (int): The number of inputs to use for the interpretation.
         use_vocab (bool): If True, the interpretation will be computed from the vocabulary of the model.
 
-    Examples:
+    Examples:  # TODO: adapt example to added arguments
```
fanny-jourdan, line 71: Didn't you adapt the example?

interpreto/concepts/interpretations/base.py
```diff
@@ -130,15 +134,33 @@ def concepts_activations_from_source(
...
+
+                if hasattr(batch_latent_activations, "to"):
+                    batch_latent_activations = batch_latent_activations.to(self.device)  # type: ignore
+                batch_concepts_activations = self.concept_model.encode(batch_latent_activations)
+
+                # SAEs outputs from overcomplete are tuples, we need to get the codes
+                if isinstance(batch_concepts_activations, tuple):
+                    batch_concepts_activations = batch_concepts_activations[1]  # temporary fix, issue #65
+
+                concepts_activations_list.append(batch_concepts_activations.cpu())  # type: ignore
```
fanny-jourdan, line 155: Do you have to have everything in the CPU to do that?

interpreto/concepts/methods/overcomplete.py
```diff
@@ -262,6 +265,9 @@ def fit(
             train_method = oc_sae.train_sae
         log = train_method(**train_params)
         self.concept_model.fitted = True
+
+        if hasattr(self.concept_model, "training"):
+            self.concept_model.training = False
```
fanny-jourdan, line 313: I don't understand. concept_model.training is an argument that doesn't exist by default, which appears when you do .fit, and now you make it appear even when you don't have .fit by setting it to False? If that's the case, why don't you set it to False in the default arguments and just set it to True when you do .fit?

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -178,7 +180,10 @@ def __init__(
             tokenizer (PreTrainedTokenizer): Custom tokenizer for the loaded model.
                 If not specified, it will be instantiated with the default tokenizer for the model.
             batch_size (int): Batch size for the model.
-            device_map (str | None): Device map for the model. Directly passed to the model.
+            device_map (torch.device | str | None): Device map for the model. Directly passed to the model.
+            output_tuple_index (int | None): If the output at the split point is a tuple, this is the index of the hidden state.
+                If `None`, an element with 3 dimensions is searched for. If not found, an error is raised. If several elements are found,
+                an error is raised.
```
fanny-jourdan: I don't understand what that is exactly. I don't understand what that is exactly. Why would it be a tuple? There could be more than two split points?

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -320,25 +326,25 @@ def _apply_selection_strategy(
         inputs: BatchEncoding | torch.Tensor,
         activations: Float[torch.Tensor, "n l d"],
         activation_granularity: ActivationGranularity,
-        aggregation_strategy: AggregationProtocol,
-    ) -> torch.Tensor:
+        aggregation_strategy: GranularityGranularityAggregationStrategy | None,
```
fanny-jourdan: I think there's a typo there. It can't be called "GranularityGranularity".

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -320,25 +326,25 @@ def _apply_selection_strategy(
...
                 In the case of a `torch.Tensor`, we assume a batch dimension and token ids.
             activations (InterventionProxy): Activations to apply selection strategy to.
             activation_granularity (ActivationGranularity): Selection strategy to apply. see :meth:`ModelWithSplitPoints.get_activations`.
-            aggregation_strategy (AggregationProtocol): Aggregation strategy to apply. see :meth:`ModelWithSplitPoints.get_activations`.
+            aggregation_strategy (GranularityGranularityAggregationStrategy | None): Aggregation strategy to apply. see :meth:`ModelWithSplitPoints.get_activations`.
 
         Returns:
-            torch.Tensor: The aggregated activations.
+            torch.Tensor: The aggregated activations. (n, l, d) or (ng, d)
+            list[list[list[int]]]: The indices of the granularity level, might be None.
```
fanny-jourdan: It's not clear to me. What exactly are these indices?

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -405,21 +421,138 @@ def _apply_selection_strategy(
...
+        new_activations: Float[torch.Tensor, "n l d"] | Float[torch.Tensor, "ng d"],
+        activation_granularity: ActivationGranularity,
+        aggregation_strategy: GranularityAggregationStrategy | None,
+        granularity_indices: list[list[list[int]]] | None = None,
+    ):
+        """
+        Reintegrates the selected activations into the initial activations.
+
+        Args:
+            initial_activations (Float[torch.Tensor, "n l d"]): The initial activations tensor.
```
fanny-jourdan, line 626: initial_activations doesn't always have that format, does it? If we take the layer just before classification in the case of task classification?

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -405,21 +421,138 @@ def _apply_selection_strategy(
...
+    def _reintegrate_selected_activations(
+        self,
+        initial_activations: Float[torch.Tensor, "n l d"],
+        new_activations: Float[torch.Tensor, "n l d"] | Float[torch.Tensor, "ng d"],
+        activation_granularity: ActivationGranularity,
+        aggregation_strategy: GranularityAggregationStrategy | None,
+        granularity_indices: list[list[list[int]]] | None = None,
+    ):
+        """
+        Reintegrates the selected activations into the initial activations.
```
fanny-jourdan, line 614: The description is not very precise.

fanny-jourdan: Actually, even after reading the entire function, I still don't understand what you're doing.

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -405,21 +421,138 @@ def _apply_selection_strategy(
...
+                return initial_activations
+            case ActivationGranularity.SAMPLE:
+                raise ValueError(
+                    "Activations aggregated at the sample level cannot be reintegrated. "
+                    "Please choose another granularity level, such as ALL_TOKENS, TOKEN, WORD, or SENTENCE."
+                )
             case _:
                 raise ValueError(f"Invalid activation selection strategy: {activation_granularity}")
 
+    def _manage_output_tuple(self, activations: torch.Tensor | tuple[torch.Tensor], split_point: str) -> torch.Tensor:
```
fanny-jourdan, line 707: Same for this function

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -474,13 +607,19 @@ def get_activations(
                     signed_max([[-1, 0, 1, 2], [-3, 1, -2, 0]]) = [-3, 1, -2, 2]
 
             pad_side (str): 'left' or 'right' — side on which to apply padding along dim=1 only for ALL strategy.
+            tqdm_bar (bool): Whether to display a progress bar.
+            return_predictions (bool): Whether to return the model predictions along with the activations.
```
fanny-jourdan: You mention the argument return_predictions, but it does not exist in the function arguments.

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -499,11 +638,209 @@ def get_activations(
             )
 
         activations: dict = {}
-        for split_point in self.split_points:
+        for split_point in self.split_points + ["predictions"]:
```
fanny-jourdan, line 908: why this add?

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -499,11 +638,209 @@ def get_activations(
...
+                    # adapt predictions ot match the granularity indices
+                    if indices_list is None:
+                        if activation_granularity == ActivationGranularity.ALL_TOKENS:
+                            activations["predictions"][-1] = (
+                                activations["predictions"][-1].repeat(1, batch_sp_activations.shape[1]).flatten(0, 1)
+                            )
+                    else:
+                        predictions = []
+                        for i, indices in enumerate(indices_list):
+                            predictions.append(activations["predictions"][-1][i].repeat(len(indices)))
```
fanny-jourdan: I still don't understand these indices_lists.

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -499,11 +638,209 @@ def get_activations(
...
+                Options are:
+                - ``ModelWithSplitPoints.activation_granularities.ALL``:
+                    the activations are returned as is ``(batch, seq_len, d_model)``.
+                    They are padded manually so that each batch of activations can be concatenated.
+
+                - ``ModelWithSplitPoints.activation_granularities.CLS_TOKEN``:
+                    only the first token (e.g. ``[CLS]``) activation is returned ``(batch, d_model)``.
+
+                - ``ModelWithSplitPoints.activation_granularities.ALL_TOKENS``:
+                    every token activation is treated as a separate element ``(batch x seq_len, d_model)``.
```
fanny-jourdan: specify whether or not you keep the padding tokens

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -499,11 +638,209 @@ def get_activations(
...
+                )
+        return activations  # type: ignore
+
+    def get_concepts_output_gradients(
+        self,
+        inputs: list[str] | torch.Tensor | BatchEncoding,
+        encode_activations: Callable[[LatentActivations], ConceptsActivations],
+        decode_activations: Callable[[ConceptsActivations], LatentActivations],
+        targets: list[int] | None = None,
+        split_point: str | None = None,
```
fanny-jourdan, line 1037: Why does the split_point argument appear here but not in get_activation?

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -523,65 +860,94 @@ def get_activations(
...
-                        activations[split_point].append(raw_activations)
-
-                        # Early stopping at the last splitting layer
-                        if idx == len(self.split_points) - 1:
-                            getattr(curr_module, module_out_name).stop()
+            # call model forward pass and gradients on concept activations
+            with self.trace(tokenized_inputs, **kwargs):
+                curr_module = self.get(local_split_point)
+                # Handle case in which module has .output attribute, and .nns_output gets overridden instead
+                module_out_name = "nns_output" if hasattr(curr_module, "nns_output") else "output"
```
fanny-jourdan: I dont understand this

interpreto/commons/granularity.py
AntoninPoche: I modified it to include the `unfold` part in a similar manner to `aggregate`.

I had a version in `model_with_split_points.py` before I merged with main. Then I tried to merge the two.

In the last commits, I modified back to `max(...).values`

interpreto/commons/granularity.py
AntoninPoche: Is this not the function to aggregate gradient-based methods?

I'm not sure what you're referring to.

interpreto/concepts/base.py
AntoninPoche: It was not the correct way to do it.

With the previous version, when initializing and user `super().__init__()` in children classes, it would have overwritten the value.

interpreto/concepts/interpretations/base.py
AntoninPoche: It is a list, so elements can be on any device, even different ones at this point. But later they are stacked, hence they need to be on the same device in the end.

Furthermore, we put things to the CPU because leaving elements on the GPU would accumulate GPU memory and end with an out of memory error.

interpreto/concepts/interpretations/topk_inputs.py
AntoninPoche: I adapted one line of the example due to a global replacement of `model_autoclass` by `automodel` to match the name of the NNsight arguments.

However, the reason for this TODO has not yet been solved.

In any case, the update of `TopKInputs` as we discussed it, will not be in this PR.

interpreto/concepts/methods/overcomplete.py
AntoninPoche: The `training` attribute is specific to a subset of overcomplete SAEs, potentially, only `BatchTopKSAE`.

The encoding forward of `BatchTopKSAE` is entirely different depending on this parameter, it was particularly a problem with incomplete batches.

Therefore, I manually set it to `False` when the training is completed because Overcomplete does not manage it.

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: I added a longer description, I hope it is better now.

In some, it is the opposite as `_apply_selection_strategy`, it is needed to go from concepts to output.

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: Well, that is what I assumed till now.

What you say is that we may have a tensor of shape `(n, l)`, because at some point, only the CLS token is left in the forward pass.

If so, can you affirm that every 2 dimensional activation tensor can be transformed into the format we expect with `(n, 1, l)`.

Otherwise, I have no idea how we may support this...

In any case, this will raise an error catched in lines 715 in `_manage_output_tuple`

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: We may want to include predictions in the final activations. (Specific to classification and depends on the `include_predicted_classes` parameter).

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: In `get_activations`, the activations for all split points of the `ModelWithSplitPoints` object are computed.

In the case of `_get_concept_output_gradients`, the `encode_activations` and `decode_activations` parameters are specific to a given `split_point`. Therefore, it should be specified.

In inherits from the duality of `ModelWithSplitPoints` which as two uses:
- get any activations we want (many split points)
- serve as a base for concept explainers (with a single split point)

In any case, this is not supposed to be called directly by the user. He should use the concept explainer `concept_output_gradient`.

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: I hope the new docstring is clearer.

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: In the middle of a model, the output of a module (for example, an attention block), the output might not be a `torch.Tensor`. If not, it is a `tuple[torch.Tensor]`. I do not know why, it does not come from `ModelWithSplitPoints` nor NNsight. It is linked with the implementation of models in `transformers`.

This is super common, as BERT and DistilBERT do so, for example.

The attribute is there for people to specify which element of the tuple we should use. If not provided, we try our luck and search for the first tensor with three dimensions.

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: This indices are the same we defined in `Granularity`, lets take an example with the `WORD` granularity.

```
example = [
    "A BC DEF",
    "abc de f"
]

indices = Granularity.get_indices(example, WORD)
# [
#      [ [0], [1, 2], [3, 4, 5] ],
#      [ [0, 1, 2], [3, 4], [5] ],
# ]
```

Here, the word `"abc"` belongs to the second sample; therefore, we need to look at the second element of the `indices` list. `"abc"` is the first word, thus the first granular element of this second sample. Therefore, `[0, 1, 2]`, which tells us that the word `"abc"` is formed with the first three tokens of the second sample.

If we had to use this information to obtain the activations of the word `"abc"`, we would look at the activations of shape (n, l, d). Then extract the elements of interest by `activations[1, [0, 1, 2],:]`, the second sample, first three tokens, all of the model dimensions. The final step would be the aggregation over the token dimension.

By applying this operation to all words, we would obtain six activation vectors, as we have six words. These are stacked to build the granular activations of shape `(ng, d)` where `ng` is the total number of granular elements over the n provided samples.

I included this explanation in the code of `_apply_selection_strategy()`.

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: I hope the example I added in `_apply_selection_strategy` helps to clarify it.

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: This is an NNsight corner case.

Usually, to access a module's output, we do `activations = path.to.module.output` in an `nnsight` `model.trace()` context.

However, ironically, some `transformers` models have modules called `output`, which leads to two different things having the same name.

NNsight chose to access the module's output in case of ambiguity by using `activations = path.to.module.nns_output`.

I do not know why they did not use the `nns_output` in all cases, but that's about it.

By the way, the line is an if-else statement one-liner. If this does not make sense, you can look it up.

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -473,121 +793,466 @@ def get_activations(
...
+
+                    # TODO: find a way to compute gradients for all targets simultaneously
+
+                    # ----------------------------------------------------------------------------------
+                    # compute gradients for each target separately
+                    targets_gradients_list: list[Float[torch.Tensor, ng, c]] = []
+                    for t in current_targets:
+                        # sum over samples but compute the gradients for each target separately
+                        with logits[t].backward(retain_graph=True):  # type: ignore
+                            # compute the gradient of the concept activations
```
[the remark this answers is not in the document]
AntoninPoche: That would be correct with PyTorch, but we use `nnsight` here, which wraps around `torch`.

The context for the backward pass is their suggested way of obtaining gradients since version 0.5.0.

tests/concepts/methods/test_sklearn_wrappers.py
```diff
@@ -0,0 +1,79 @@
...
+
+    sk_wrapper.to(DEVICE)
+
+    # Compare encodings
+    encoded_wrapper = sk_wrapper.encode(data)
+    encoded_sklearn = torch.as_tensor(sk_model.transform(data.cpu().numpy()), dtype=torch.float32)
+
+    assert isinstance(encoded_wrapper, torch.Tensor)
+    assert encoded_wrapper.shape == encoded_sklearn.shape == torch.Size([n, c])
+    assert torch.allclose(encoded_wrapper.cpu(), encoded_sklearn, atol=1e-5, rtol=1e-5)
```
fanny-jourdan: I prefer when the assert returns a comment so that we know immediately what the problem is.

tests/concepts/methods/test_concept_autoencoder_explainers.py
```diff
@@ -141,6 +142,74 @@ def test_overcomplete_cbe(
...
+    if not cbe.has_differentiable_concept_decoder:
+        pytest.skip("Skipping test for method_class that does not have a differentiable concept decoder")
+
+    gradients = cbe.concept_output_gradient(
+        sentences,
+        targets=[0],
+        activation_granularity=granularity,
+    )
+    assert isinstance(gradients, list)
+    assert len(gradients) == len(sentences)
```
fanny-jourdan: Same for the asserts on this file

tests/concepts/methods/test_concept_autoencoder_explainers.py
```diff
@@ -141,6 +142,74 @@ def test_overcomplete_cbe(
...
+            padding=True,
+            truncation=True,
+            return_offsets_mapping=True,
+        )
+        indices_list = Granularity.get_indices(tokens, granularity.value, tokenizer)  # type: ignore
+        nb_granularity_elements = len(indices_list[0])
+        assert grad.shape == (1, nb_granularity_elements, concepts_dim)
+
+
 if __name__ == "__main__":
```
fanny-jourdan, line 244: why you use a main and not directly @pytest.parametrize ?

tests/model_wrapping/test_model_with_split_points.py
```diff
@@ -32,23 +34,43 @@
...
 ]
 
+GRANULARITIES = [
+    ActivationGranularity.ALL,
+    ActivationGranularity.ALL_TOKENS,
+    ActivationGranularity.CLS_TOKEN,
+    ActivationGranularity.TOKEN,
+    ActivationGranularity.WORD,
+    ActivationGranularity.SENTENCE,
+    ActivationGranularity.SAMPLE,
```
fanny-jourdan, line 62: what is the difference between ALL, ALL_TOKENS and SAMPLE ?

tests/model_wrapping/test_model_with_split_points.py
```diff
@@ -86,111 +106,361 @@ def test_loading_possibilities(bert_model, bert_tokenizer, gpt2_model, gpt2_toke
...
+def test_pad_and_concat():
+    """Validate ``pad_and_concat`` left and right padding."""
+    tensors = [
+        torch.zeros(1, 2, 3),
+        torch.ones(1, 3, 3),
+    ]
+    out_right = MWSP._pad_and_concat(tensors, "right", 0.5)
+    out_left = MWSP._pad_and_concat(tensors, "left", -1.0)
+
+    assert out_right.shape == (2, 3, 3)
```
fanny-jourdan: Same for the asserts of this files

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -24,42 +24,38 @@
...
-)
-from interpreto.typing import LatentActivations
+from interpreto.typing import ConceptsActivations, LatentActivations
 
 
 class InitializationError(ValueError):
     """Raised to signal a problem with model initialization."""
 
 
 class ActivationGranularity(Enum):
```
fanny-jourdan, line 55: Can you give a definition of each granularity on this class?

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -68,33 +64,107 @@ class ActivationGranularity(Enum):
...
 
+AG = ActivationGranularity
+
+
 class ModelWithSplitPoints(LanguageModel):
     """Code: [:octicons-mark-github-24: model_wrapping/model_with_split_points.py` ](https://github.com/FOR-sight-ai/interpreto/blob/dev/interpreto/model_wrapping/model_with_split_points.py)
 
-    Generalized NNsight.LanguageModel wrapper around encoder-only, decoder-only and encoder-decoder language models.
-    Handles splitting model at specified locations and activation extraction.
+    The `ModelWithSplitPoints` is a wrapper around your HuggingFace model.
```
fanny-jourdan, line 109: Super clear docs, thanks !

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -103,139 +173,120 @@ class ModelWithSplitPoints(LanguageModel):
...
         self.batch_size = batch_size
 
         if not isinstance(model_or_repo_id, str):
-            if device_map == "auto":
-                raise ValueError(
-                    "'auto' device_map is only supported when loading a model from a repository id. "
-                    "Please specify a device_map, e.g. 'cuda' or 'cpu'."
-                )
-            self.to(device_map)
+            if device_map is not None:
```
fanny-jourdan, line 316: What do you do when device_map is None?

tests/concepts/methods/test_concept_autoencoder_explainers.py
AntoninPoche: The main part is only used for debugging; it does not influence the pytest pipeline.

I cannot use VSCode debugger with pytest, hence I write an `if __name__ == "__main__":` section and run the specific tests with the debugger.

I could remove these lines when I do not need them anymore. But I often found them useful afterward.

tests/model_wrapping/test_model_with_split_points.py
AntoninPoche: Well, I should add this description in the `ActivationGranularity` Enum! But they are described in the documentation of `MWSP.get_activations()`.

In some:
- `ALL`: the raw activations (n, l, d)
- `ALL_TOKENS`, a flattened version of `ALL`, (n*l, d)
- `SAMPLE` activations aggregated for all relevant tokens. Take the `TOKENS` activations and aggregate using the `aggregation_strategy` over all the tokens of a given sample. Final shape: (n, d)

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: Sure, I will do it! Though it will mainly be a copy from `MWSP.get_activations()` docstring.

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -473,121 +793,466 @@ def get_activations(
...
                 names in `self.split_points`, while values correspond to the extracted activations for the split point
                 for the given `inputs`.
         """
-        if not self.split_points:
-            raise RuntimeError(
-                "No split points are currently defined for the model. "
-                "Please set split points before calling get_activations."
-            )
+        # set default pad side value and catch unsupported cases
+        if self._model.__class__.__name__.endswith("Classification"):
```
fanny-jourdan: I'm not sure that just putting “Classification” is enough. Maybe “sequenceclassification” (I don't have the exact spelling) would be better, because then the user could give a Token Classification model.

interpreto/model_wrapping/model_with_split_points.py
```diff
@@ -473,121 +793,466 @@ def get_activations(
...
+            pad_side = pad_side or "right"
```
fanny-jourdan: Maybe we can force the use of right padding for classification and left padding for generation? Because if the user puts something else, it doesn't make much sense to me.

interpreto/concepts/methods/sklearn_wrappers.py
```diff
@@ -0,0 +1,409 @@
...
+        self.fitted = True
+
+        if return_sklearn_model:
+            return ica
+
+        del ica
+
+    @jaxtyped(typechecker=beartype)
+    def encode(self, x: Float[torch.Tensor, "n {self.input_size}"]) -> Float[torch.Tensor, "n {self.nb_concepts}"]:
+        self._assert_fitted()
```
fanny-jourdan, line 130: Does self._assert_fitted() return an error if you haven't done the fit before? I imagine that's how it works, but I'm asking just in case.

interpreto/concepts/methods/sklearn_wrappers.py
```diff
@@ -0,0 +1,409 @@
...
+
+class SkLearnWrapper(BaseOptimDictionaryLearning):
+    def __init__(
+        self,
+        nb_concepts: int,
+        input_size: int,
+        *,
+        random_state: int = 0,
+        device: torch.device | str = "cpu",
+    ):
```
fanny-jourdan, line 60: It would be nice to have a description of the arguments here or in each subclass (whichever you prefer).

interpreto/concepts/methods/sklearn_wrappers.py
```diff
@@ -0,0 +1,409 @@
...
+
+    def get_dictionary(self):
+        self._assert_fitted()
+        return self.components
+
+
+_SkLearnWrapper_co = TypeVar("_SkLearnWrapper_co", bound=SkLearnWrapper, covariant=True)
+
+
+class SkLearnWrapperExplainer(DictionaryLearningExplainer[SkLearnWrapper], Generic[_SkLearnWrapper_co]):
```
fanny-jourdan, line 310: I dont understand what (DictionaryLearningExplainer[SkLearnWrapper], Generic[_SkLearnWrapper_co]) mean for create the class. Exactly what does it herit?

fanny-jourdan (approved):
Thank you, Antonin, for this very comprehensive PR. The documentation is particularly clear, which I really appreciate. I have a few minor questions and comments that I will leave you to address, but overall everything looks good.

interpreto/concepts/methods/sklearn_wrappers.py
AntoninPoche: Yes, that is the goal of this function.

interpreto/concepts/methods/sklearn_wrappers.py
AntoninPoche: First, this is mainly typing and does not significantly impact the behavior. It is to please the linter.

If we remove the typing part, what's under the hood is:
```python
class SkLearnWrapperExplainer(DictionaryLearningExplainer[SkLearnWrapper], Generic[_SkLearnWrapper_co]):
class SkLearnWrapperExplainer(DictionaryLearningExplainer):
```

Now let's dive into the `Generic` design pattern:

We initially define `ConceptEncoderExplainer` with a `concept_model` attribute of type `ConceptModel`. The type `ConceptModel` tells you that it should follow the `ConceptModelProtocol` (have a `nb_concepts`, can be `.fit()`, and `.encode()`).

Apart from the protocol, nothing is exceptional so far (I can detail this too if needed). But what we want is for the type of the `concept_model` to define the type of the concept explainer. For example, if your `concept_model` is an autoencoder and not an encoder, then you have a `ConceptAutoEncoderExplainer` and not a `ConceptEncoderDecoder`. However, we write this the other way around, saying the `concept_model` of a `ConceptEncoderDecoder` is a `ConceptModel`, and for `ConceptAutoEncoderExplainer`, we have a `BaseDictionaryLearning`.

If we had only these two classes, it would be completely overkill. But we do this for all the inheriting classes, an `NMFConcepts` explainer has an `NMF` `concept_model`. 

Finally, how do we specify this to Python? In the parent class, we say it is a `Generic` over the `concept_model` type induced by this class, and in the children class, we add `[specific concept model type]`. Here is an example with three layers of inheritance to understand it. Because the generic typing is not inherited, it needs to be respecified. Note that the linter is happy only if the children class `concept_model` type inherits from the parent class specified `concept_model` type in the `Generic`.

```python
class GrandParent(Generic[GrandParentConcept]):
    pass

class Parent(GrandParent[ParentConcept], Generic[ParentConcept]):
    pass

class Child(Parent[ChildConcept]):
    pass
```

Ultimately, it allows us to specify the type of the `concept_model` via inheritance. I know it is not easy to read. But after searching for a long time, it was the cleanest typing I found.

interpreto/model_wrapping/model_with_split_points.py
AntoninPoche: Two variables affect the behavior here: the type of `model_or_repo_id` and `device_map`.

If `model_or_repo_id` is a string, then `transformers` loads it. The default `device_map` value is `None`, it loads the model on the PU.

If `model_or_repo_id` is a model, then `nnsight` ignores the `device_map` parameter. It is our job to manage it. **If it is `None`, we leave the model on its device.** Otherwise, we put the given model to the specified `device_map`.

I wrote comments to clarify this.

tests/model_wrapping/test_model_with_split_points.py
AntoninPoche: The docstring was clarified.

PR #73, opened 2025-07-31
Introduce Insertion and Deletion metrics for attribution methods

cofri:
## Description

This PR introduces the Insertion and Deletion metrics for attribution methods. It includes the following new features:
- new `InsertionPerturbator` and `DeletionPerturbator` that perturbs the inputs based on the order of importance of the tokens. The importance is given by the attribution scores. The Perturbators return masks where elements are iteratively inserted/removed based on the importance.
- new `InsertionDeletionAggregator` that reformats scores obtained from the perturbed sequences in order to compute the metric AUC.
- the new metrics: `Insertion` and `Deletion`. They are built as an AttributionExplainer with custom perturbator and aggregator (defined above). These new classes return an average AUC score and the raw scores for each sequence with the successive perturbations. The raw scores are useful to plot the curve on which the AUC score is computed.
- Unit tests are added for these new classes.

For a seamless integration, two changes are proposed:
- the `AttributionOutput` was extended to store the "model_inputs_to_explain", the targets, the granularity and the inference mode.
- the base `Perturbator` now accepts kwargs that are transferred to the `get_mask` method. This is required to give the attribution outputs to `get_mask` in order to compute the mask based on the importance of the attributions.

Finally, some bugs are fixed:
- in attribution methods, `tokenizer` is replaced with `self.tokenizer` to avoid confusion on possible different objects.
- in Occlusion method, `replace_token_id` argument was removed since it is now retrieved from the tokenizer, as done in other methods. The behaviour is now unified across methods.

## Type of Change

## Checklist

### DIFF
--- a/interpreto/attributions/aggregations/insertion_deletion_aggregation.py
+++ b/interpreto/attributions/aggregations/insertion_deletion_aggregation.py
@@ -1,0 +1,94 @@
+# MIT License
+#
+# Copyright (c) 2025 IRT Antoine de Saint Exupéry et Université Paul Sabatier Toulouse III - All
+# rights reserved. DEEL and FOR are research programs operated by IVADO, IRT Saint Exupéry,
+# CRIAQ and ANITI - https://www.deel.ai/.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Aggregator for Insertion and Deletion metrics
+"""
+
+import torch
+from jaxtyping import Float
+from torch import Tensor
+
+from interpreto.attributions.aggregations.base import Aggregator
+
+
+class InsertionDeletionAggregator(Aggregator):
+    """
+    Aggregator for Insertion and Deletion metrics.
+
+    This class aggregates the results of the insertion and deletion metrics by filtering the results based on the
+    number of perturbations per target. The results are expected to be in the shape of (p * t, t), where p is the
+    number of perturbations per target and t is the number of targets. The aggregation is done by reshaping the results
+    to (p, t) where each row corresponds to a target and each column corresponds to a perturbation. For each target,
+    the results are aggregated by taking the values corresponding to the perturbations for that target.
+
+    Example:
+    Below is an example with t=3 targets (columns) and p=2 perturbations per target (rows). The interesting values are
+    represented by 1, 2, and 3, for each target. The "*" values are not considered in the aggregation. The aggregation
+    only keeps the scores of the perturbations corresponding to each target, resulting in a tensor of shape (p, t).
+    ```
+    results =
+    [
+        1a * *
+        1b * *
+        * 2a *
+        * 2b *
+        * * 3a
+        * * 3b
+    ]
+
+    aggregation =
+    [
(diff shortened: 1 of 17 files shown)


cofri:
I've just added a new commit a9882f4 to support NaN values in the attributions. NaN values correspond to future tokens in generation tasks.
This new commit updates the mask generation (`InsertionPerturbator` and `DeletionPerturbator`), and adds a unit test with NaN values.

interpreto/attributions/methods/kernel_shap.py
```diff
@@ -98,7 +98,7 @@ def __init__(
         model, replace_token_id = self._set_tokenizer(model, tokenizer)
 
         perturbator = ShapTokenPerturbator(
-            tokenizer=tokenizer,
+            tokenizer=self.tokenizer,
```
fanny-jourdan, line 101: Why add a self.tokenizer here and in all other inference-based methods rather than just tokenizer? We don't need self since we give the argument directly from init without doing “self.tokenize.”

AntoninPoche: The tokenizer might be modified in `self._set_tokenizer` by adding the `[PAD]` token into the vocabulary.

Therefore, the perturbator need to have this new tokenizer which is saved in `self.tokenizer`.

It worked before because `tokenizer` and `self.tokenizer` are two pointers to the same object. Nonetheless, the correct way to write it is to add the `self`.

tests/attributions/metrics/test_insertion_deletion.py
fanny-jourdan, line 1: Shouldn't we also test for generation task?

cofri: I've added a test on a generation task (commit 7bdb14a4613dba811301bdb2303f2cc6974fdadc)

docs/notebooks/attribution_walkthrough.ipynb
AntoninPoche, line 1: You may not have modified the notebook. I do not know if it is easy to remove it from the PR. (If not, ignore this point.)

However, it could be great to include metrics in the walkthrough.

interpreto/attributions/aggregations/insertion_deletion_aggregation.py
```diff
@@ -0,0 +1,94 @@
...
+        * 2b *
+        * * 3a
+        * * 3b
+    ]
+
+    aggregation =
+    [
+        1a 2a 3a
+        1b 2b 3b
+    ]
```
AntoninPoche, line 64: Not optimal because it calls the get_scores for all targets while you only need one for each perturbation.

Here is what you might need in the Base explainer:

```
    def target_generator(
        self, perturbations: Iterable[TensorMapping], targets: torch.Tensor
    ) -> Iterable[tuple[TensorMapping, torch.Tensor]]:
        for pert, target in zip(perturbations, targets, strict=True):
            total_pert = pert["input_ids"].shape[0]
            assert total_pert % len(target) == 0
            n_pert = total_pert // len(target)
            yield pert, target.repeat_interleaves(n_pert)  # [0, 0, ..., 1, 1, ..., 2, 2, ...]
```

AntoninPoche: Then this class might not be necessary as the reshaping is much simpler.

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -0,0 +1,272 @@
...
+            for score, mask in zip(scores, mask_generator, strict=True)
+        )
+
+        # Compute AUC using trapezoidal rule
+        cleaned_scores = [s.cpu().numpy() for s in cleaned_scores]
+        aucs = []
+        for s in cleaned_scores:
+            auc = ((s.sum(axis=0) - 0.5 * (s[0] + s[-1])) / (len(s) - 1)).mean()  # Average AUC across targets
+            aucs.append(auc)
+        auc = float(np.mean(aucs))
```
AntoninPoche, line 175: Is this not what we expect of the aggregator?

Returning both the auc and scores is not a problem.

interpreto/attributions/perturbations/insertion_deletion_perturbation.py
```diff
@@ -0,0 +1,148 @@
...
+        most_important_idx = torch.argsort(attributions_inf, descending=True)
+
+        # Compute the real number of perturbations "p": take the minimum between n_perturbations and the number of
+        # elements to perturb (+1 for the baseline)
+        num_elements_to_perturb = int(mask_dim * self.max_percentage_perturbed)
+        p = min(self.n_perturbations, max(num_elements_to_perturb, 1)) + 1
+
+        # Build the mask
+        mask: Float[torch.Tensor, "p l"] = self._baseline_mask((p * num_targets, mask_dim))
+        steps = np.linspace(0, max(1, num_elements_to_perturb), p)
```
AntoninPoche, line 134: It would be better to stay in PyTorch.

interpreto/attributions/perturbations/insertion_deletion_perturbation.py
```diff
@@ -0,0 +1,148 @@
...
+        fill_value = int(1 - mask[0, 0])  # Value to fill the mask with (1 for deletion, 0 for insertion)
+        for j in range(num_targets):
+            for i, step in enumerate(steps):
+                mask[(p * j) + i, most_important_idx[j, : int(step)]] = fill_value
```
AntoninPoche, line 138: I suggest that you first build a tensor of shape (t, p, l) of perturbed elements and then reshape it to (t * p, l) with `view`.

This would allow for tensorial indexing (at least I think). Furthermore, it would be much easier to get back to the desired shape for aggregation.

PR #88, opened 2025-09-10
Logit lens

Raphael-Bernas:
## Description

Implemented Logit Lens

## Type of Change

## Checklist


FETCH_HEAD
AntoninPoche, line 1: Please remove this file or tell me its use.

docs/api/others/methods/logit_lens.md
```diff
@@ -0,0 +1,55 @@
...
+::: interpreto.others.logit_lens_general.BaseLogitLens
+    handler: python
+    options:
+      show_root_heading: true
+      show_source: true
+      inherited_members: true
+      members:
+        - do_lens
+        - explain
+        - __call__
```
AntoninPoche: Do the base classes add any information compared to the classification and generation one for the user?

For me, you can leave only the classes used by the user.

docs/api/others/overview.md
AntoninPoche, line 1: The user wants all the information to be at the same place. Please merge this file into the LogitLens classes docstrings.

docs/notebooks/other_methods_walkthrough.ipynb
AntoninPoche, line 1: Please rename this file `logit_lens_example.ipynb`.

It should be compiled so users can see the result and what it looks like before trying it themselves.

I tried running it, but it did not compile. Similarly to the tests, there is a problem with `model_autoclass` -> `automodel`.

I did not analyze the notebook as I do not have the outputs.

interpreto/others/logit_lens_general.py
AntoninPoche, line 1: You can simply call this file `logit_lens.py`.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        tokenizer: PreTrainedTokenizer,
+        head_name: str | None = None,
+        model_head: nn.Module | None = None,
+        features_dim: int | None = None,
+        normalization: bool = False,
+        normalization_method: nn.Module | None = nn.LayerNorm,
+        device: torch.device | None = None,
+        batch_size: int = 8,
+    ):
+        # Check for meta tensors and reload model directly if needed
```
AntoninPoche: This lacks the attribute descriptions.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            # Re-initialize model head
+            self.head_name = head_name
+            self._initialize_head_and_features(features_dim)
+
+        if self.needs_model_reload(verbose=True):
+            raise RuntimeError(
+                "Failed to access model head even after reloading the model. Please check the model and head configuration."
+            )
+
+    def needs_model_reload(self, verbose: bool = False) -> bool:
```
AntoninPoche: If methods are not supposed to be used by users, please set them to protected or private.

If this is supposed to be used, the docstring is insufficient.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                self.model = AutoModelForCausalLM.from_pretrained(model_path, device_map=target_device)
+            elif model_autoclass == modeling_auto.AutoModelForMaskedLM:
+                self.model = AutoModelForMaskedLM.from_pretrained(model_path, device_map=target_device)
+            elif model_autoclass == modeling_auto.AutoModelForSequenceClassification:
+                self.model = AutoModelForSequenceClassification.from_pretrained(model_path, device_map=target_device)
+            else:
+                # Fallback to AutoModel for other cases
+                self.model = AutoModel.from_pretrained(model_path, device_map=target_device)
+
+            print(f"✅ Successfully reloaded model directly with proper weights on {target_device}.")
```
AntoninPoche: The code should not include print functions.

If you want to do logging, please use logging tools, set to quiet by default.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            H = model_head(test_tensor)
+            if torch.all(H < 1e-14) and torch.all(H > -1e-14):
+                if verbose:
+                    print("Model head returned all-zero output, indicating potential meta tensor issue.")
+                return True  # Head likely has meta tensors
+            return False  # Head works fine
+        except Exception as e:
+            if verbose:
+                print(f"Model head access failed: {e}")
+            return True  # Head access failed
```
AntoninPoche: This test does not convince me.

What are meta-tensors? Please explain the problem in more detail.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+    def _get_model_or_repo_id(self) -> str:
+        """
+        Retrieve the original model_or_repo_id parameter.
+
+        Returns:
+            str: The repository ID or model path
+        """
+        # Try to get repo_id from the model
+        if hasattr(self.splitted_model, "repo_id") and self.splitted_model.repo_id:
+            return self.splitted_model.repo_id
```
AntoninPoche: I do not think the model is ever stocked as `self.splitted_model` in `ModelWithSplitPoints`.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            # Determine the appropriate model class and load directly
+            model_autoclass = self.splitted_model.model_autoclass
+            target_device = self.device
+
+            if model_autoclass == modeling_auto.AutoModelForCausalLM:
+                self.model = AutoModelForCausalLM.from_pretrained(model_path, device_map=target_device)
+            elif model_autoclass == modeling_auto.AutoModelForMaskedLM:
+                self.model = AutoModelForMaskedLM.from_pretrained(model_path, device_map=target_device)
+            elif model_autoclass == modeling_auto.AutoModelForSequenceClassification:
+                self.model = AutoModelForSequenceClassification.from_pretrained(model_path, device_map=target_device)
```
AntoninPoche: While working with Interpreto and nnsight, I never saw anything related to "meta tensors." When I checked, it could be seen as a device with only the architecture loaded.

You getting the the repo id and loading back the model can cause far too many problems.

If the meta tensors are indeed a problem (which I am unsure).

We should be able to prevent it before end. I prefer the user to have an error raised and reload the model himself rather than loading it ourselfs like this.

interpreto/others/logit_lens_general.py
AntoninPoche: I think you can just check the device of the model or weights, and it should tell you if you have tensors on the meat device, right?

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        # Check main model
+        try:
+            for param in self.model.parameters():
+                if param.device.type == "meta":
+                    return True
+        except Exception:
+            print("Warning: Could not check for meta tensors.")
+            pass
+
+        return False
```
AntoninPoche: Method not used in the code.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        if self.head_name is None:
+            print("No head name specified, trying to find a suitable head in the model.")
+            for head_name in possible_heads:
+                try:
+                    head = self.model
+                    for part in head_name.split("."):
+                        head = getattr(head, part)
+
+                    if isinstance(head, nn.Linear):
+                        self.model_head = head
```
AntoninPoche: Few things here:
- First, it is hard to judge the code when you do not have comments about the expected behavior.
- Second, you might find so many corner cases by going through layer names to try your luck, that I can find cases where your code does not extract the head you wanted.
Finally, for things like this, I suggest forcing the user to provide the head's name himself. Make sure to test it and guide him if the head name does not work.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                            break
+                except AttributeError:
+                    continue
+        elif hasattr(self.model, self.head_name):
+            self.model_head = getattr(self.model, self.head_name)
+            k += 1
+            if not isinstance(self.model_head, nn.Module):
+                raise ValueError(f"The pre-set head '{self.head_name}' is not a valid nn.Module.")
+        else:
+            raise ValueError(f"The specified head '{self.head_name}' does not exist in the model.")
```
AntoninPoche: Listing the possibilities might be great here.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            )
+        elif k > 1:
+            raise ValueError(
+                f"Multiple heads found in the model. Please specify a valid head name. Found: {k} possible heads."
+            )
+
+        # Determine features dimension
+        if features_dim is not None:
+            self.features_dim = features_dim
+            print(f"Using manually specified features dimension: {self.features_dim}")
```
AntoninPoche: What do you do if the user does not specify the features_dim corresponding to the model?

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            elif hasattr(self.model_head, "decoder") and hasattr(self.model_head.decoder, "in_features"):
+                return self.model_head.decoder.in_features
+            elif hasattr(self.model_head, "predictions") and hasattr(self.model_head.predictions, "decoder"):
+                if hasattr(self.model_head.predictions.decoder, "in_features"):
+                    return self.model_head.predictions.decoder.in_features
+            else:
+                for _name, module in self.model_head.named_modules():
+                    if isinstance(module, nn.Linear) and hasattr(module, "in_features"):
+                        return module.in_features
+        return None
```
AntoninPoche: This is not the correct way to look at the problem.

You will never be able to list all the names of the submodules that might the `in_features` attributes. Worse, it might not be the first submodule that you find.

What you can do is to use the `get_latent_shape` method from `ModelWithSplitPoints`.

Or recursively check the first submodule for its `in_features`.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                              nb_token=6,
+                              batch_size=2)
+
+        # Analyze how predictions evolve across layers
+        logit_lens.lens("The cat sat on the")
+
+        # For classification models
+        cls_lens = LogitLens(classification_model, tokenizer,
+                           pooling_strategy="cls")
+        cls_lens.lens("This movie is great!")
```
AntoninPoche: This does not appear correctly in the documentation. Take example on other files, you should have `    >>> your_code`.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        Returns:
+            Processed logits tensor
+        """
+        pass
+
+    def _explain(
+        self, inputs: str | list[str] | BatchEncoding | torch.Tensor, layers_name: str | list[str] | None = None
+    ):
+        """Generate explanations using the Logit Lens method."""
+        if isinstance(inputs, (str, list)):
```
AntoninPoche: `isinstance(inputs, (str, list))` -> `isinstance(inputs, (str | list))` as suggested by the `make lint` errors.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        self.activations = self.splitted_model.get_activations(
+            inputs_model, ModelWithSplitPoints.activation_granularities.ALL
+        )
+
+        if layers_name is None:
+            layers_name = list(self.activations.keys())
+
+        for layer in layers_name:
+            if layer not in self.activations:
+                raise ValueError(f"Layer '{layer}' not found in the model activations.")
```
AntoninPoche: This test should be done prior to calling explanations. It could save the user a lot of time.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        # Validate activations
+        self._validate_activations(layers_name)
+        if isinstance(layers_name, str):
+            layers_name = [layers_name]
+
+        self.layer_names = layers_name
```
AntoninPoche: This can surely be included in the `self._validate_activations` method.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        """Safely extract item from tensor, handling meta tensors."""
+        if tensor.device.type == "meta":
+            raise RuntimeError("Cannot extract item from meta tensor. Please ensure your model is properly loaded.")
+        return tensor.detach().cpu().item()
+
+    def _safe_tensor_to_cpu(self, tensor):
+        """Safely move tensor to CPU, handling meta tensors."""
+        if tensor.device.type == "meta":
+            raise RuntimeError("Cannot move meta tensor to CPU. Please ensure your model is properly loaded.")
+        return tensor.detach().cpu()
```
AntoninPoche: Might not be necessary anymore if we solve the meta tensor problem.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                )
+                for i in range(0, input_ids.shape[0], self.batch_size)
+            ]
+        elif isinstance(inputs, torch.Tensor):
+            batched_inputs = torch.split(inputs, self.batch_size, dim=0)
+        else:
+            raise ValueError("Unsupported type for inputs")
+
+        if layers_name is None:
+            print("Producing lens prediction for all activations.")
```
AntoninPoche: Remove prints in all your files.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                BatchEncoding(
+                    {
+                        "input_ids": input_ids[i : i + self.batch_size],
+                        "attention_mask": attention_mask[i : i + self.batch_size],
+                    }
+                )
+                for i in range(0, input_ids.shape[0], self.batch_size)
+            ]
+        elif isinstance(inputs, torch.Tensor):
+            batched_inputs = torch.split(inputs, self.batch_size, dim=0)
```
AntoninPoche: Why not let `ModelWithSplitPoints.get_activations()` handle the batching?

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+
+        merged_results = {}
+        for batch in batched_inputs:
+            batch_results = self._explain(batch, layers_name)
+            for layer, data in batch_results.items():
+                if layer not in merged_results:
+                    merged_results[layer] = data
+                else:
+                    # Merge batch results
+                    self._merge_batch_results(merged_results[layer], data)
```
AntoninPoche: Be careful, here, for each batch, you concatenate and create a new tensor. This is not good practice for memory usage and computation time.

You should make a list of data for a given layer, then concatenate them all.

AntoninPoche (changes requested):

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+    Logit Lens implementation for Language Models (Causal LM and Masked LM).
+
+    This implementation shows what tokens the model "thinks" should come next
+    or fill a mask at each layer.
+    """
+
+    def __init__(self, model: ModelWithSplitPoints, tokenizer: PreTrainedTokenizer, nb_token: int = 5, **kwargs):
+        self.vocab_size = tokenizer.vocab_size if tokenizer else None
+        self.nb_token = min(nb_token, self.vocab_size) if self.vocab_size else nb_token
+        super().__init__(model, tokenizer, **kwargs)
```
AntoninPoche: There should be a description of each attribute and parameter.

What is `nb_token` used for? How does it differ from `vocab_size`?

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        """Enable the instance to be called as a function."""
+        return self.explain(inputs, layers_name)
+
+
+class LanguageModelLogitLens(BaseLogitLens):
+    """
+    Logit Lens implementation for Language Models (Causal LM and Masked LM).
+
+    This implementation shows what tokens the model "thinks" should come next
+    or fill a mask at each layer.
```
AntoninPoche: The description should be much more precise and concrete. Users might not know what a logit lens is.

Furthermore, I would add a warning in the docstring that the activation distribution shifts with layers. Hence, the predicted tokens might not make sense in the early layer due to this distribution shift.

Ultimately, I suggest copying most of the docstring between the classes the user might use for the logit lens.

It would be best if the user only uses one class.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        Apply the Logit Lens method to the model activations.
+
+        Args:
+            activation: Model activations tensor
+            inputs_model: Optional model inputs for accessing attention mask
+
+        Returns:
+            Processed logits tensor
+        """
+        pass
```
AntoninPoche: It seems that this method should not be used by the user, hence, it should be protected.

Furthermore, the name is not precise at all. I suggest calling it `_predict_logits_from_activations`.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                )
+
+            # Convert to probabilities
+            proba = F.softmax(logits, dim=-1)
+
+            # Get predicted classes
+            predicted_classes = torch.argmax(proba, dim=-1)
+
+            # Get top-k predictions
+            top_k_values, top_k_indices = torch.topk(proba, k=min(self.num_classes, 5), dim=-1)
```
AntoninPoche: The argmax is included in the topk. You do the topk line and write: `predicted_classes = top_k_indices[:, 0]`.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                    [
+                        [
+                            proba[batch_idx, seq_idx, top_k_indices[batch_idx, seq_idx]].detach().cpu().numpy()
+                            for seq_idx in range(top_k_indices.shape[1])
+                        ]
+                        for batch_idx in range(top_k_indices.shape[0])
+                    ]
+                ),
+            }
+        return results
```
AntoninPoche: This seemed really unoptimal to me. I asked our friend ChatGPT, and it gave me the following, which seems much better:

```
def process_logits(logits_dict, tokenizer, nb_token: int):
    results = {}

    for layer, logits in logits_dict.items():
        # Compute probabilities
        proba = F.softmax(logits, dim=-1)

        # Get top-k indices and values directly (avoids recomputing probabilities later)
        top_k_values, top_k_indices = torch.topk(proba, nb_token, dim=-1)

        # Flatten indices for fast batch decoding
        flat_indices = top_k_indices.reshape(-1).tolist()

        # Batch decode once instead of looping per token
        decoded_tokens = tokenizer.convert_ids_to_tokens(flat_indices)

        # Reshape to original [batch, seq, k]
        decoded_tokens = np.array(decoded_tokens, dtype=object).reshape(
            top_k_indices.shape
        )

        # Move proba to numpy (already aligned with indices)
        top_k_proba = top_k_values.detach().cpu().numpy()

        results[layer] = {
            "tokens": decoded_tokens,
            "proba": top_k_proba,
        }

    return results
```

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        if isinstance(layers_name, str):
+            layers_name = [layers_name]
+
+        self.layer_names = layers_name
+
+        # Process activations through the lens
+        logits_dict = {}
+        for layer in layers_name:
+            logits_dict[layer] = self.do_lens(self.activations[layer], inputs_model)
+        return self._process_logits(logits_dict, inputs_model)
```
AntoninPoche: Here, you loop on layers, then call the `self.process_logits`, which also loops on layers. I suggest that your `_process_logits` only treats the logits for a single layer, and you loop on them here.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            if activation.shape[-1] != self.features_dim:
+                raise ValueError(
+                    f"Activation shape mismatch for layer '{layer}': expected {self.features_dim}, got {activation.shape[-1]}."
+                )
+
+    def _process_logits(self, logits_dict: dict[str, torch.Tensor], inputs_model) -> dict:
+        """Process logits for language modeling - convert to probabilities and get top-k tokens."""
+        if self.nb_token == 0:
+            print("WARNING: `nb_token` is set to 0, no top-k tokens will be returned.")
+            return {layer: F.softmax(logits, dim=-1) for layer, logits in logits_dict.items()}
```
AntoninPoche: For me, it does not make sense to set `nb_tokens` to zero. It should raise an error at initialization. Similarly, when you have a larger `nb_tokens` than the `vocab_size`, it should raise an error.

Ho and `nb_token` should be plural: `nb_tokens`.

interpreto/others/logit_lens_general.py
AntoninPoche: This is really similar to what is done with generation:
- softmax on dim -1
- topk
- convert indices to strings
- make an array of probabilities and an array of strings

Therefore, it should be factorized, and the difference should be in the conversion between indices and strings.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                "probabilities": proba.detach().cpu().numpy(),
+                "predicted_classes": predicted_classes.detach().cpu().numpy(),
+                "predicted_labels": [
+                    class_labels.get(self._safe_item(idx), f"Class_{self._safe_item(idx)}")
+                    for idx in predicted_classes
+                ],
+                "top_k_predictions": top_k_predictions,
+                "class_labels": class_labels,
+                "confidence_scores": torch.max(proba, dim=-1)[0].detach().cpu().numpy(),
+            }
```
AntoninPoche: I am sure there is much more information than you need here!

Some are duplicated (proba include confidence scores).

Why do you need logits? You already have proba.
Why do you need class IDs? You already have their class name.
Is the label not similar to the class name?

Finally, please don't use the term probabilities (in your all file); these are not calibrated. The logits after softmax can be called scores.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            return {0: "Negative", 1: "Positive"}
+        elif self.num_classes == 3:
+            # Common for sentiment analysis
+            return {0: "Negative", 1: "Neutral", 2: "Positive"}
+        elif self.num_classes == 5:
+            # Common for rating/sentiment (1-5 stars)
+            return {i: f"{i + 1}_star" for i in range(5)}
+        else:
+            # Generic labels as fallback
+            return {i: f"Class_{i}" for i in range(self.num_classes)}
```
AntoninPoche: I do not think we should try to automatically extract the class names, just ask the user to provide them or write `Class_0`, `Class_1`, ...

There are too many cases where this will not do what we expect. Please keep it simple.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                existing_data[key] = np.concatenate([existing_data[key], value], axis=0)
+            # For any other data types, try to extend if it's a list
+            elif isinstance(existing_data[key], list):
+                if isinstance(value, list):
+                    existing_data[key].extend(value)
+                else:
+                    existing_data[key].append(value)
+            else:
+                # If we can't merge, keep the existing data
+                pass
```
AntoninPoche: If you factorize most of the `_process_logits` functions, you should have similar outputs for both classification and generation, and a much simpler merging.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                np.pad(
+                    existing_data["proba"], ((0, 0), (0, existing_pad), (0, 0)), mode="constant", constant_values=0.0
+                ),
+                np.pad(new_data["proba"], ((0, 0), (0, new_pad), (0, 0)), mode="constant", constant_values=0.0),
+            ],
+            axis=0,
+        )
+
+        existing_data["tokens"] = merged_tokens
+        existing_data["proba"] = merged_proba
```
AntoninPoche: I do not think you need to do padding. You will not have to pass your logit lens output tokens through a model.

Therefore, it can be stored in a list with an element for each sample.

The merging thus becomes an append to a list.

This can surely be generalized to the classification part.

interpreto/others/logit_lens_general.py
AntoninPoche: I said in a previous comment that this function should treat only one layer at a time.

Therefore, the output does not have to be a dict of dict. But it can be a tuple of tensors (tokens, proba). This would be much better for type checking and clarity.

It is better to refrain from using dict for outputs when you can.

interpreto/others/logit_lens_general.py
AntoninPoche: In fact, the conversion from indices to strings can even be left to later. It can be seen as a visualization problem. Indices take less space than strings.

Furthermore, it allows you to stay with torch tensors the whole time.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                        # Special tokens
+                        grouped.append(f"⟨{token[1:-1]}⟩")
+                    else:
+                        current_word_parts = [token]
+
+            # Don't forget the last word
+            if current_word_parts:
+                grouped.append("".join(current_word_parts))
+
+            return grouped
```
AntoninPoche: Never called.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            cleaned = cleaned.replace("▁", " ")
+
+            # End-of-word markers
+            cleaned = cleaned.replace("</w>", "")
+
+            # BERT-style: ## prefix indicates continuation of previous word
+            if cleaned.startswith("##"):
+                cleaned = cleaned[2:]  # Remove ## prefix
+
+            return cleaned
```
AntoninPoche: Utils functions should be defined outside of the classes.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            Clean up tokenization artifacts to make tokens more readable.
+
+            Args:
+                token: Raw token from tokenizer
+
+            Returns:
+                Cleaned token with artifacts removed
+            """
+            # Handle different tokenization schemes
+            cleaned = token
```
AntoninPoche: not super useful.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                ),
+                np.pad(new_data["proba"], ((0, 0), (0, new_pad), (0, 0)), mode="constant", constant_values=0.0),
+            ],
+            axis=0,
+        )
+
+        existing_data["tokens"] = merged_tokens
+        existing_data["proba"] = merged_proba
+
+    def visualize_logit_lens_interactive( # noqa: PLR0912  # ignore too many branches  # too many special cases
```
AntoninPoche: Your notebooks are not compiled, hence I cannot evaluate your visualization.

In any case, this function should be placed in a visualization-related file. You should not need the Logit lens method inside it. Please refactor your code so that your visualization function takes in the output of your logit lens method.

For inputs, you can always convert the input token IDs to tokens.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                        current_word_parts = [token]
+
+            # Don't forget the last word
+            if current_word_parts:
+                grouped.append("".join(current_word_parts))
+
+            return grouped
+
+        explaining = self.explain(inputs, layers_name)
+        layers_name = self.layer_names if layers_name is None else layers_name
```
AntoninPoche: do not compute your method inside the visualization

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+
+        if self.nb_token == 0:
+            print(
+                "WARNING: `nb_token` is set to 0, no top-k tokens will be returned. If you want to see the top-k tokens, please set `nb_token` to a positive integer."
+            )
+            return
```
AntoninPoche: I said it in a previous comment, but raise an error at the initialization, please.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        explaining = self.explain(inputs, layers_name)
+        layers_name = self.layer_names if layers_name is None else layers_name
+
+        if self.nb_token == 0:
+            print(
+                "WARNING: `nb_token` is set to 0, no top-k tokens will be returned. If you want to see the top-k tokens, please set `nb_token` to a positive integer."
+            )
+            return explaining
+
+        self.visualize_logit_lens_interactive(inputs, layers_name)
```
AntoninPoche: I think we should discuss the API. We have a function that calls everything, or we split the computations and the plotting.

I argue for the latter; it is common practice to do so. And you can add many more visualization parameters. Finally, some people may want the output without the visualization.

AntoninPoche: Here, you compute the explanations, but you do not give them to the visualization nor return them.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            inputs = inputs.unsqueeze(0).unsqueeze(0)
+
+        explaining = self.explain(inputs, layers_name)
+        layers_name = self.layer_names if layers_name is None else layers_name
+
+        if self.nb_token == 0:
+            print(
+                "WARNING: `nb_token` is set to 0, no top-k tokens will be returned. If you want to see the top-k tokens, please set `nb_token` to a positive integer."
+            )
+            return explaining
```
AntoninPoche: Same `nb_tokens <= 0` comment.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+
+    This implementation shows what classification decision the model is leaning
+    towards at each layer.
+    """
+
+    def __init__(
+        self,
+        model: ModelWithSplitPoints,
+        tokenizer: PreTrainedTokenizer,
+        pooling_strategy: str = "cls",  # "cls", "mean", "last"
```
AntoninPoche: Please rely on the `GranularityAggregationStrategy` if you need several strategies. But I think classification should force the cls token.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        **kwargs,
+    ):
+        self.pooling_strategy = pooling_strategy
+        self.original_pooling_strategy = pooling_strategy  # Keep track of original strategy
```
AntoninPoche: Why are there two different?

interpreto/others/logit_lens_general.py
AntoninPoche: You are lacking the arguments description and example.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            ]
+            if len(linear_modules) == 1:
+                # Only one linear layer, use its out_features
+                return linear_modules[0].out_features
+            elif len(linear_modules) > 1:
+                # More than one linear layer, pick the last one (usually the classifier output)
+                return linear_modules[-1].out_features
+        raise ValueError(
+            "Could not determine number of classes from the model head. Please set the number of classes manually using `set_num_classes`."
+        )
```
AntoninPoche: You do not need to bother extracting the number of classes from here. You can trust the shape of the logits.

But if you ask for the class names, you can check that the logits shape is coherent with the number of classes.

You were also using this attribute to extract the class names automatically, but you can check my comment there.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+
+    def set_num_classes(self, num_classes: int):
+        """Set the number of classes for the classification head."""
+        self.num_classes = num_classes
```
AntoninPoche: `num_classes` is neither protected nor a property; it does not need a setter. Furthermore, this attribute is not useful.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+
+    def _detect_and_handle_head_pooling(self):
+        """
+        Detect if the model head performs its own pooling and adjust the pooling strategy accordingly.
+        This handles cases like RobertaClassificationHead that expect full sequences and pool internally.
+        """
+        if not self.model_head:
+            return
```
AntoninPoche: This should never be the case; a `LogitLens` object without a `model_head` makes no sense.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+    This class defines the common interface and shared functionality for different
+    types of logit lens implementations (Language Model vs Classification).
+    """
+
+    def __init__(
+        self,
+        model: ModelWithSplitPoints,
+        tokenizer: PreTrainedTokenizer,
+        head_name: str | None = None,
+        model_head: nn.Module | None = None,
```
AntoninPoche: The model head it key to the `LogitLens` object it should be provided.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+                len(head_modules) > 2 and self.pooling_strategy is not None
+            ):  # More than just the head itself and one linear layer
+                print(
+                    f"⚠️  Warning: {self.model_head.__class__.__name__} appears to be a complex head with multiple layers."
+                )
+                print(f"   Current pooling strategy: '{self.pooling_strategy}'")
+                print(
+                    "   If you encounter dimension mismatches, the head might be incompatible with external pooling."
+                )
+                print("   Consider setting pooling_strategy=None if issues arise.")
```
AntoninPoche: This is clearly LLM-generated. It is not a bad thing to use LLMs, but do you know what this does?

It tries to find out the expected shape from the name of the classification head module and submodules. As I said in several prior comments. This is not what we should be doing. You should force the user to set a pooling parameter.

AntoninPoche: Therefore, remove this function.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        elif self.pooling_strategy == "mean":
+            # Use mean pooling
+            if attention_mask is not None:
+                mask = attention_mask.unsqueeze(-1).expand(activation.size())
+                masked_activation = activation * mask
+                return masked_activation.sum(dim=1) / mask.sum(dim=1)
+            else:
+                return activation.mean(dim=1)
+        else:
+            raise ValueError(f"Unknown pooling strategy: {self.pooling_strategy}")
```
AntoninPoche: Well, you can do this by setting the `activation_granularity` parameter of `ModelWithSplitPoints.get_activations()`. You can save a lot of memory and some lines of code. (None: `ALL`, "cls": `CLS_TOKEN`, "last": `LAST_TOKEN`, "mean": `SAMPLE`).

The `LAST_TOKEN` is not yet implemented. But if you really need it quickly I can do it.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+
+        if self.normalization:
+            if input_to_head.dim() == 3:
+                # Apply normalization to the last dimension for 3D tensors
+                norm_layer = self.normalization_method(input_to_head.size(-1))
+                input_to_head = norm_layer(input_to_head)
+            else:
+                # Apply normalization for 2D tensors
+                norm_layer = self.normalization_method(input_to_head.size(-1))
+                input_to_head = norm_layer(input_to_head)
```
AntoninPoche: duplicated code.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        if layers_name is None:
+            layers_name = list(self.activations.keys())
+
+        for layer in layers_name:
+            if layer not in self.activations:
+                raise ValueError(f"Layer '{layer}' not found in the model activations.")
+        # Validate activations
+        self._validate_activations(layers_name)
+        if isinstance(layers_name, str):
+            layers_name = [layers_name]
```
AntoninPoche: In `_validate_activations`, you iterate on the layers, which you do many time. I am sure, you can drastically reduce the number of loops.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        Validate activations for classification models.
+        """
+        for layer in layers_name:
+            activation = self.activations[layer]
+            if activation.dim() not in [2, 3]:
+                raise ValueError(
+                    f"Activation for layer '{layer}' must be 2D (batch_size, hidden_dim) or 3D (batch_size, seq_len, hidden_dim), got shape: {activation.shape}"
+                )
+
+            # For 3D activations, we can handle any hidden dimension via pooling
```
AntoninPoche: I do not agree, for Roberta, this would be a problem.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+            # For 2D activations, they should match the model head's expected input dimension
+            if activation.dim() == 2 and activation.shape[-1] != self.features_dim:
+                raise ValueError(
+                    f"2D activation for layer '{layer}' should match features_dim: expected {self.features_dim}, got {activation.shape[-1]}"
+                )
```
AntoninPoche: I am quite sure this type of check are done at the end of the `ModelWithSplitPoints.get_activations()`, in any cases, it was extensively tested. You can have confidence in this.

I do not think you need this method.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+    3. Converting the resulting logits to probabilities to see what the model would
+       predict if it stopped processing at that layer
+    4. Visualizing how predictions change and confidence builds up across layers
+
+    Originally introduced in "Interpreting GPT: the logit lens" by nostalgebraist:
+    https://www.lesswrong.com/posts/AcKRB8wDpdaN6v6ru/interpreting-gpt-the-logit-lens
+
+    Automatically chooses the appropriate implementation based on the model type:
+    - **Language Models** (AutoModelForCausalLM, AutoModelForMaskedLM): Uses LanguageModelLogitLens
+    - **Classification Models** (AutoModelForSequenceClassification): Uses ClassificationLogitLens
```
AntoninPoche: When you make lists in markdown, you should leave a line before the bullet points or numbers; otherwise, it does not render.

You can do `make docs` to check how your documentation renders.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+    Args:
+        model: The wrapped model to analyze
+        tokenizer: The tokenizer corresponding to the model
+        **kwargs: Additional arguments passed to the specific LogitLens implementation
+
+    Returns:
+        Appropriate LogitLens implementation (LanguageModelLogitLens or ClassificationLogitLens)
```
AntoninPoche: Not necessary.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+    Originally introduced in "Interpreting GPT: the logit lens" by nostalgebraist:
+    https://www.lesswrong.com/posts/AcKRB8wDpdaN6v6ru/interpreting-gpt-the-logit-lens
+
+    Automatically chooses the appropriate implementation based on the model type:
+    - **Language Models** (AutoModelForCausalLM, AutoModelForMaskedLM): Uses LanguageModelLogitLens
+    - **Classification Models** (AutoModelForSequenceClassification): Uses ClassificationLogitLens
+    Args:
+        model: The wrapped model to analyze
+        tokenizer: The tokenizer corresponding to the model
+        **kwargs: Additional arguments passed to the specific LogitLens implementation
```
AntoninPoche: The arguments do not appear in the documentation because you have not `__init__` method. This is normal because you have a factory. But we should find a way for them to appear in the documentation.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        cls_lens = LogitLens(classification_model, tokenizer,
+                           pooling_strategy="cls")
+        cls_lens.lens("This movie is great!")
+    """
+
+    # Mapping of model types to their corresponding LogitLens implementations
+    _model_type_to_lens = {
+        # Language Models
+        modeling_auto.AutoModelForCausalLM: LanguageModelLogitLens,
+        modeling_auto.AutoModelForMaskedLM: LanguageModelLogitLens,
```
AntoninPoche: Did you test it? This might not work directly. If not, a new class might be needed.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+
+        html += "</div>"
+
+        display(HTML(html))
+
+    def lens(self, inputs: str | list[str] | BatchEncoding | torch.Tensor, layers_name: str | list[str] | None = None):
+        """
+        Convenient method to visualize classification lens results.
+        """
+        self.visualize_classification_lens(inputs, layers_name)
```
AntoninPoche: I do not like this; there is no documentation and it is just an alias.

Well I made my comments on the generation part of this.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+        model_class_name = model._model.__class__.__name__
+
+        if any(name in model_class_name.lower() for name in ["classification", "classifier"]):
+            print(f"Detected classification model from class name: {model_class_name}")
+            print("Using ClassificationLogitLens")
+            return ClassificationLogitLens(model, tokenizer, **kwargs)
+        elif any(name in model_class_name.lower() for name in ["causal", "lm", "masked", "language"]):
+            print(f"Detected language model from class name: {model_class_name}")
+            print("Using LanguageModelLogitLens")
+            return LanguageModelLogitLens(model, tokenizer, **kwargs)
```
AntoninPoche: Remove prints.

interpreto/others/logit_lens_general.py
```diff
@@ -0,0 +1,1420 @@
...
+    def lens(self, inputs: str | list[str] | BatchEncoding | torch.Tensor, layers_name: str | list[str] | None = None):
+        """
+        Convenient method to visualize classification lens results.
+        """
+        self.visualize_classification_lens(inputs, layers_name)
+
+
+class LogitLens:
+    """
+    Factory class for creating LogitLens implementations.
```
AntoninPoche: After reading the whole code, you might be able to do a single class for classification and generation if you extract the visualization.

Most specific methods involve automating things from the module names, but you can also leave the user to do most of it.

Otherwise, the key difference was `_process_logits`, but if you return topk token IDs or class IDs, I think you can manage.

My suggestion is to:
- remove the automatization from names,
- factorize things into the base class iteratively
- extract the visualization

If there is still code in the specific classes, we can discuss it at this moment.

logit_lens.ipynb
AntoninPoche, line 1: Please remove this file from the git.

mkdocs.yml
```diff
@@ -11,6 +11,7 @@ nav:
   - Tutorials:
     - Use Attribution Methods: notebooks/attribution_walkthrough.ipynb
     - Training a Concept Explainer: notebooks/concept_examples.ipynb
+    - Other Interpretability Methods: notebooks/other_methods_walkthrough.ipynb
```
AntoninPoche: Name it Logit Lens; otherwise, no one will find it.

mkdocs.yml
```diff
@@ -43,6 +44,10 @@ nav:
         - Dictionary Metrics: api/concepts/metrics/dictionary_metrics.md
         - Reconstruction Metrics: api/concepts/metrics/reconstruction_metrics.md
         - Sparsity: api/concepts/metrics/sparsity_metrics.md
+    - Other Methods:
+      - Overview: api/others/overview.md
+      - Methods:
+        - Logit Lens: api/others/methods/logit_lens.md
```
AntoninPoche: After looking at the documentation, I think people will find it easier if you just put `Logit Lens` at the same level as attributions and concepts (in the documentation).

Then, you can merge the overview file into the logit lens classes docstrings. There should be a single documentation file left.

Please refer to the `ModelWithSplitPoints` documentation in your classes documentation. (I do not remember how, maybe `:class:~ModelWithSplitPoints`).

tests/others/test_logit_lens.py
```diff
@@ -0,0 +1,1226 @@
...
+}
+
+CLASSIFICATION_MODELS = {
+    "hf-internal-testing/tiny-random-bert": {
+        "model_class": AutoModelForSequenceClassification,
+        "split_points": ["bert.encoder.layer.1.output"],
+        "head_name": "classifier",
+        "pooling_strategy": "cls",
+    },
+}
```
AntoninPoche: `hf-internal-testing` does not have initialized classification heads. That may explain your meta tensor error.

tests/others/test_logit_lens.py
```diff
@@ -0,0 +1,1226 @@
...
+
+CI_MODELS = {
+    "causal_lm": ["hf-internal-testing/tiny-random-gpt2"],
+    "masked_lm": ["hf-internal-testing/tiny-random-bert"],
+    "classification": ["hf-internal-testing/tiny-random-bert"],
+}
```
AntoninPoche: This is not enough model diversity.

tests/others/test_logit_lens.py
```diff
@@ -0,0 +1,1226 @@
...
+
+# Test fixtures
+@pytest.fixture
+def sentences():
+    """Test sentences for experiments."""
+    return [
+        "Interpreto is the latin for 'to interpret'. But it also sounds like a spell from the Harry Potter books.",
+        "Interpreto is magical",
+        "Testing interpreto",
+    ]
```
AntoninPoche: You do not have to define it, it already exist in `tests/conftest.py`, therefore it is automatically detected as a fixture.

tests/others/test_logit_lens.py
AntoninPoche, line 1: Most of your tests concern methods that automatically set parameters from module names. I suggested you remove these parts, so you can remove the tests too.

You tried several models, but it is far from enough due to the diversity of how models manage their heads.

I assume that with more models, the methods above would break.

I suggested too many modifications for the tests to remain relevant next time.

I will review the whole code again next time.

AntoninPoche (changes requested):
Hi @Raphael-Bernas, I copied my previous review comment as most of my remarks are still relevant.

You will see the comments, but I basically told you to rework a bit of everything. If you think my comments are harsh, please do not take it to heart, I am focused on the code while reviewing. There are things to improve but you did a great job.



Note that everything that I say can be false; I insist, do not hesitate to challenge my comments. I do not have the complete picture of your code as you do. Sometimes I understand things later. 


I suggested too many changes, so you can work on the same branch, I will review the whole code again, do not worry.



Here are a few first general comments:
- You tried to treat too many cases with your first version. You iterated on your code without it being reviewed. You might have had fewer things to modify if you had interacted more with us. (But it is also our fault). 
- Several of your choices are not trivial (e.g., a factory or reloading the model). It is better to discuss architectural decisions like these with others before committing to the idea. Because they impact the final API, it is a team choice. Furthermore, we might have elements to help you as we have already developed other parts.
- You must pass the `make lint` and `make test` before asking for reviews. The test on the PR's GitHub page should also pass before asking for a review, but you can ask for help.
- You need to install the Pyright package in your editor and set the checks to "standard." This checks that the types you specified are coherent, preventing many uncaught bugs. You should not have red underlines from the type checker; otherwise, there is something to modify. You can say `# type: ignore`, but this should not be the default behavior.
- Do not leave `print` in your code, not your tests. You can add logs if you really need to.
- We need to discuss the meta tensor problem. We might be able to prevent it, but we should not load the model in the LogitLens init.
- Several points can be managed by `ModelWithSplitPoints`.
- The code lacks many comments and docstrings. These are super important, first for the users in the documentation, then for reviewers, and finally, for maintainers.

More precise comments now:
- You can simplify your doc by merging the overview file into the classes' docstrings.
- You wrote many functions to automatically adapt the code's behavior to the module names. There might be as many possibilities as models; hence, you cannot treat them all. It is not a problem to force the user to provide some parameters and guide them through documentation and error raising.
- I think the documentation should be in a different file, but we can discuss.
- After removing the automation methods and visualization, I think that most of the code can be factored into the base class. Hence, there might be a possibility that we only need one class. Which would be much more suitable for documentation and users.



I know it is a lot of remarks. It would have been easier if you had submitted something smaller or discussed more. I am sorry I did not catch it earlier.

PR #89, opened 2025-09-17
ConSim metric

AntoninPoche:
## Description

Adds the `ConSim` metric for concept-based explainers as introduced by:
A. Poché, A. Jacovi, A.M. Picard, V. Boutin, and F. Jourdan. [ConSim: Measuring Concept-Based Explanations' Effectiveness with Automated Simulatability](https://aclanthology.org/2025.acl-long.279/). In the Proceedings of the 2025 Association for Computational Linguistics (ACL).

I recommend making the doc before and going through the documentation for an easier dive into the method.

## Type of Change

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "interpreto"
-version = "0.4.2"
+version = "0.4.3"
 description = "Interpretability toolbox for LLMs"
 readme = "README.md"
 requires-python = ">=3.10"


interpreto/concepts/metrics/consim.py
```diff
@@ -24,54 +24,1300 @@
...
+            [c[:nb_mistakes_elements_per_class] for c in class_wise_incorrect_indices]
+        )[:nb_mistakes]
+
+        # in case the number of correct or incorrect is not a multiple of the number of classes
+        nb_correct_remaining = nb_correct - nb_correct_elements_per_class * nb_classes
+        if nb_correct_remaining:
+            additional_possible_correct_indices = torch.cat(
+                [c[nb_correct_elements_per_class:] for c in class_wise_correct_indices]
+            )
+            new_indices = torch.randint(len(additional_possible_correct_indices), (nb_correct_remaining,))
```
AntoninPoche, line 445: Codex: [P1] Guard remainder sampling when classes exceed requested samples

If nb_lp_samples + nb_ep_samples is smaller than twice the number of classes, the warning branch forces nb_correct_elements_per_class to 1. In that case nb_correct_remaining becomes negative, yet the truthy check on line 435 still executes and calls torch.randint with a negative shape on line 439, raising a runtime error instead of returning a sample set. This crashes select_examples for small evaluation batches. The remainder blocks should only run when the counts are strictly positive.

interpreto/concepts/metrics/consim.py
```diff
@@ -24,54 +24,1300 @@
...
+                Each element of the list if a dictionary with an importance associated to a concept id.
+                Filtered to keep only the important concepts.
+        """
+        # ------------------------------------------------------------------------------------------
+        # filter concepts which are important for at least one class
+        concepts_to_keep = []
+        while len(concepts_to_keep) == 0:
+            for concepts_importance in global_importances.values():
+                if len(concepts_importance) == 0:
+                    continue
```
AntoninPoche, line 625: Codex: [P1] Prevent infinite loop when global concept importances sum to zero

The loop that selects `concepts_to_keep` halves the threshold until some concept exceeds it, but when all per-class dictionaries are empty or their absolute importances sum to zero, `normalized_importances` becomes NaN and `concepts_to_keep` never fills. The loop at line 616 therefore runs forever, repeatedly halving the threshold and hanging the evaluation. A break or default fallback is needed when no valid importances exist.

interpreto/concepts/metrics/consim.py
```diff
@@ -24,54 +24,1300 @@
...
+
+        Returns:
+            predictions: list[str] | None
+                The model predictions.
+                If the response is empty or the expected length is not respected, returns None.
+        """
+        sentences = response.split("\n")
+
+        while sentences[-1] == "":
+            sentences = sentences[:-1]
```
AntoninPoche, line 1071: Codex: [P1] Handle empty LLM responses in prediction extraction

When `user_llm_response` is an empty string, `response.split('\n')` returns `[""]`; the while loop at lines 1059‑1062 repeatedly indexes `sentences[-1]` after the list has been truncated to length 0, raising `IndexError` instead of returning `None`. This means a silent failure from the meta-predictor crashes ConSim instead of producing a score of `None`. Check that the list is non-empty before accessing `sentences[-1]` or early-return on empty responses.

interpreto/concepts/metrics/consim.py
```diff
@@ -24,54 +24,1300 @@
...
+        >>> dataset = datasets.load_dataset("fancyzhx/ag_news")
+        >>> classes = ["World", "Sports", "Business", "Sci/Tech"]
+        >>> activations = model_with_split_points.get_activations(dataset["train"]["text"])
+        >>>
+        >>> # -------------------------
+        >>> # Fit the concept explainer
+        >>> concept_explainer_1 = ICAConcepts(model_with_split_points, nb_concepts=50)
+        >>> concept_explainer.fit(activations)
+
+        The two steps of ConSim:
```
fanny-jourdan, line 241: The example is very clear, thank you. But in practice, I think we would need a metric notebook on the doc in addition to our attribution and concept notebooks, where you would show how to use ConSim in a use case.

interpreto/concepts/metrics/consim.py
```diff
@@ -24,54 +24,1300 @@
...
+                The labels of the interesting samples.
+            predictions: torch.Tensor
+                The predictions of the model on the interesting samples.
+        """
+        nb_classes = len(self.classes) if self.classes is not None else len(torch.unique(labels))
+        nb_correct = (nb_lp_samples + nb_ep_samples) // 2
+        nb_mistakes = nb_lp_samples + nb_ep_samples - nb_correct
+
+        # Find the correct and incorrect indices
+        is_prediction_correct = predictions == labels
```
fanny-jourdan, line 394: I imagine you checked that it worked, but I'm just asking in case, because the notation is a bit strange without parentheses like that.

interpreto/concepts/metrics/consim.py
```diff
@@ -24,54 +24,1300 @@
...
+            seed=seed,
+        )
+
+    @staticmethod
+    def _quantize_importances(importance: float, threshold: float = 0.05) -> str | None:
+        """
+        Convert the normalized importances to literals.
+        The literals are:
+        - "++" for values above 0.3
+        - "+" for values between 0.05 and 0.3
```
fanny-jourdan, line 529: The 0.05 isn't fixed here, is it? It corresponds to the variable named threshold?  
And if that's the case, it seems strange to me that you can vary the value to get + and - but not the one to get ++ and --.

interpreto/concepts/metrics/consim.py
```diff
@@ -24,54 +24,1300 @@
...
+        The concepts importance are quantized to literals.
+        The corresponding literals are:
+
+        - "++" for values above 0.3
+
+        - "+" for values between 0.05 and 0.3
+
+        - "-" for values between -0.05 and -0.3
+
+        - "--" for values below -0.3
```
fanny-jourdan, line 582: Same here

interpreto/concepts/metrics/consim.py
```diff
@@ -24,54 +24,1300 @@
...
+                The accuracy of the model predictions.
+                If the model predictions are empty or the user-llm predictions are empty, returns None.
+
+        Raises:
+            ValueError
+                If the model predictions and the user-llm predictions have different lengths.
+        """
+        if len(model_predictions) == 0 or len(user_llm_predictions) == 0:
+            # TODO: discuss if we prefer to return None or raise an error with explicit message
+            # I argue for returning 0, as most of the problems come for the user_llm giving incoherent responses
```
fanny-jourdan: Maybe just add a warning?

interpreto/concepts/metrics/consim.py
```diff
@@ -24,54 +24,1300 @@
...
+            score or prompts and model predictions: float | None | tuple[list[tuple[Role, str]], list[str]]
+                The score of the ConSim metric.
+                If the user-llm predictions are empty or in the wrong format, returns None.
+                It was chosen to return None,
+                because ConSim should be called a lot of times for statistically significant results.
+                Therefore, having a None score once in a while is better than the script crashing.
+
+                If no user_llm is provided, returns the prompts and the model predictions.
+                The user will have to call the ConSim prompts manually.
+                The response of the LLM on the prompts should be compared to the model predictions.
```
fanny-jourdan: This is not super clear

interpreto/concepts/metrics/consim.py
AntoninPoche: Excellent point, I will do all the notebooks simultaneously in a later PR.

tests/concepts/metrics/test_consim.py
```diff
@@ -0,0 +1,716 @@
...
+        # extract the sample indices from the user prompt
+        ep_samples_str = user_prompt.split("\n\nConcepts contributions for Sample_")[0]
+        # Format:
+        #     Sample_0: "this is the first sample"
+        #     Sample_1: "this is the second sample"
+        ep_sample_indices = [int(s.split(":")[0][7:]) for s in ep_samples_str.split("\n") if s]
+
+        # generate the response, give a random class for each sample
+        response = "\n".join([f"Sample_{i}: {classes[i % len(classes)]}" for i in ep_sample_indices])
+        return response
```
fanny-jourdan, line 87: For testing, it would also be interesting to have an LLMInterfacePlaceholder2 that does not produce a very good output format (to check how ConSim behaves in such cases).

tests/concepts/metrics/test_consim.py
```diff
@@ -0,0 +1,716 @@
...
+def test_consim_init(splitted_encoder_ml: ModelWithSplitPoints, multi_split_model: ModelWithSplitPoints):
+    """
+    Test the `__init__` method of the ConSim metric.
+    """
+    llm = LLMInterfacePlaceholder()
+    classes = [str(i) for i in range(int(splitted_encoder_ml._model.num_labels))]  # type: ignore
+
+    # when only one split point is available, it should be chosen automatically
+    consim = ConSim(splitted_encoder_ml, llm, AG.TOKEN, classes=classes)
+    assert consim.split_point == splitted_encoder_ml.split_points[0]
```
fanny-jourdan: please add comment on your asserts

fanny-jourdan (approved):
Nice work, thanks Antonin ! I added a few comments, but everything seems fine to me. :)

interpreto/concepts/metrics/consim.py
AntoninPoche: I was not sure, so I tested on a small example:
```
>>> import torch
>>> p = torch.tensor([0, 1, 2, 1, 0])
>>> l = torch.tensor([0, 1, 2, 3, 4])
>>> correct = p == l
>>> print(correct)
tensor([ True,  True,  True, False, False])
```

In some, whatever you write after `x =` will be executed before assigning this value to x. The `==` between two tensors of the same shape will do a pointwise comparison and create a tensor with the shape of boolean values.

interpreto/concepts/metrics/consim.py
AntoninPoche: Indeed, there is no `0.3` in the code, but only `6 * threshold`.

I will specify in the documentation that it is an example with the default value. I think people will not move the threshold in practice, so writing them makes more sense than repeating the code in the documentation string.

Issue #96, opened 2025-10-06
Jacobian calculation taking too much GPU memory with small LLMs


ClementBM-CLB:
## Question

When applying gradient based attribution to LLMs like llama 3.2 3B or larger, the [get_gradients()](https://github.com/FOR-sight-ai/interpreto/blob/d8999a74fee514399e209df72e543db94729ac31/interpreto/model_wrapping/inference_wrapper.py#L599C9-L599C22) method from `inference_wrapper.py` take a lot of GPU RAM because the jacobian is calculated for each batch in parallel.

Would it be possible to make the calculation sequential?

## Checklist

- [x] I've searched the project's [`issues`](https://github.com/FOR-sight-ai/interpreto/issues?q=is%3Aissue).

PR #97, opened 2025-10-06
Making the jacobian calculation sequential

ClementBM-CLB:
## Description

It's just a draft to make the jacobian calculation sequential.
It would require additional effort to make this optional.

## Related Issue

#96 

## Type of Change

## Checklist

### DIFF
--- a/interpreto/model_wrapping/inference_wrapper.py
+++ b/interpreto/model_wrapping/inference_wrapper.py
@@ -271,7 +271,11 @@ class InferenceWrapper(ABC):
                 f"Consider adjust the batch size or the wrapper of split your data.",
             )
         # Check sequence length
-        if input_ids is not None and getattr(self.model.config, "max_position_embeddings", False) and input_ids.shape[-1] > self.model.config.max_position_embeddings:
+        if (
+            input_ids is not None
+            and getattr(self.model.config, "max_position_embeddings", False)
+            and input_ids.shape[-1] > self.model.config.max_position_embeddings
+        ):
             raise ValueError(
                 f"Input sequence length ({input_ids.shape[1]}) exceeds model's maximum "
                 f"input length ({self.model.config.max_position_embeddings}). Please truncate your inputs by specifying 'truncation=True' or 'max_length={self.model.config.max_position_embeddings}' to the tokenizer call or change the model."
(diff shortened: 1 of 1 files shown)


AntoninPoche:
Thank you very much for the pull request.

The current interpreto code did too many computations, creating a much larger matrix than necessary.

Your version works well, but I am looking into it to see if I can prevent the for loop with a `vmap`.

interpreto/model_wrapping/inference_wrapper.py
```diff
@@ -638,24 +642,52 @@ def _get_gradients_from_mapping(
...
+        if len(grad_matrix.shape) == 3:
+            return grad_matrix
+        else:
+            batch_size = inputs_embeds.shape[0]
+            target_size = 1
+            if len(targets.shape) == 1:
+                target_size = targets.shape[0]
+            elif len(targets.shape) == 2:
+                target_size = targets.shape[1]
+            return grad_matrix.view((batch_size, target_size, -1))
```
AntoninPoche, line 690: Well I made few modifications:

- define the `get_sample_score` in the for loop to prevent the double function definition (`get_sacore` and `lambda`).
- index the `grad_matrix_` to prevent the squeeze() and thus the if statements at the end
- Few more comments

```python
# Iterate on samples to call the Jacobian on each sample.
grad_matrices: list[torch.Tensor] = []
for i, input_embeds in enumerate(inputs_embeds):
    attention_mask = model_inputs["attention_mask"][i].unsqueeze(0)

    # take the sample-wise target if it is a batch of targets
    if len(targets.shape) == 2 and targets.shape[0] > 1:
        current_targets = targets[i].unsqueeze(0)
    else:
        current_targets = targets

    # define the function to call in the Jacobian (1, l, d) -> (1, lt) 
    def get_sample_score(inputs_embeds):
        return self.get_targeted_logits(
            {
                "inputs_embeds": inputs_embeds,
                "attention_mask": attention_mask,
            },
            current_targets,
        )
    
    # get gradient for each target
    grad_matrix_: torch.Tensor = torch.autograd.functional.jacobian(
        get_sample_score,
        input_embeds.unsqueeze(0),
    )  # (1, lt, 1, l, d)
    grad_matrices.append(grad_matrix_[:, :, 0])

grad_matrix = torch.cat(grad_matrices)  # (n, lt, l, d)

if input_x_gradient:
    grad_matrix = grad_matrix * inputs_embeds.unsqueeze(1)

grad_matrix = grad_matrix.abs().mean(
    dim=-1
)  # (n, lt, l)  # average over the embedding dimension  # TODO: discuss if the average should be forced or an argument
return grad_matrix

```

AntoninPoche (changes requested):
Thank you very much for your PR @ClementBM-CLB!

I tried the `vmap`, but I did not manage to make it work in this case.

Still, I suggested several modifications, if you are good with them, could you include them?

Otherwise, this is good to be merged.

PR #102, opened 2025-10-07
Jacobian OOM due to n x n computations of the gradient

AntoninPoche:
## Description

The Jacobian is computed on a batch. However, the Jacobian operator from torch uses computes the gradient of all inputs (n, l, d) to all outputs (n, lt). Thus, creating a tensor (n, lt, n, l, d) leads to OOM. Furthermore, this computes n x n x lt gradients compared to the n x lt required.

This is the interpreto follow-up to #97, trying to solve #96.

## Type of Change

## Checklist


interpreto/attributions/perturbations/linear_interpolation_perturbation.py
```diff
@@ -81,10 +81,14 @@ def adjust_baseline(baseline: TensorBaseline, inputs: torch.Tensor) -> torch.Ten
         # Shape: (batch_size, *input_shape)
         input_shape = inputs.shape[1:]
 
+        # When all values are zero, the gradients are always NaN.
+        # To avoid this, we set the baseline to a small value.
         if baseline is None:
```
thomas-mullor: Duplicate IF clause. I think it could be merged

interpreto/attributions/perturbations/linear_interpolation_perturbation.py
```diff
@@ -81,10 +81,14 @@ def adjust_baseline(baseline: TensorBaseline, inputs: torch.Tensor) -> torch.Ten
...
-            baseline = 0
+            baseline = 1e-6
+        if isinstance(baseline, int | float) and baseline in [0, 0.0]:
```
thomas-mullor: First condition is implied by the second one

interpreto/commons/granularity.py
```diff
@@ -302,6 +292,46 @@ def __word_get_indices(word_ids: torch.Tensor) -> list[list[int]]:
...
+            return False
+        if token.startswith("@@"):
+            return False
+        if token.startswith("__"):
+            return True
+        if token.startswith("Ġ"):
+            return True
+        if token.startswith(" "):
+            return True
+        return False
```
thomas-mullor: As Python use short circuit evaluation, these conditions could be aligned directly in the return clause without loss of efficiency. Moreover, the "False" cases are unnecessary as they check conditions to return the default result

thomas-mullor: One-liner could be something like :
`return token.startswith(" ") or token.startswith("Ġ") or token.startswith("__")`

interpreto/model_wrapping/inference_wrapper.py
```diff
@@ -584,6 +588,13 @@ def get_targeted_logits(
             f"get_targeted_logits not implemented for {self.__class__.__name__}. Implement this method is necessary to use gradient-based methods."
         )
 
+    @get_targeted_logits.register(MutableMapping)
```
thomas-mullor: This method shouldn't be defined at this level.
Since the base method is abstract, dispatching logic should be implemented only in the relevant subclasses (or not). Users may implement custom subclasses of the inference wrapper without needing to perform dispatching or naming their registration methods _get_targeted_logits_from_mapping.

interpreto/model_wrapping/inference_wrapper.py
```diff
@@ -631,31 +642,99 @@ def get_gradients(
...
-        grad_matrix = grad_matrix.abs().mean(
-            dim=-1
-        )  # (n, lt, l)  # average over the embedding dimension  # TODO: discuss if th average should be forced or an argument
-        return grad_matrix
+        inputs_embeds = model_inputs["inputs_embeds"].detach().requires_grad_(True)  # (n,l,d)
+        attention_mask = model_inputs["attention_mask"]  # (n, l)
+
+        # Compute logits for ALL classes once if that’s cheaper in your model
+        # and select later inside get_targeted_logits via gather.
+        logits = self._get_targeted_logits_from_mapping(
```
thomas-mullor: This method may not be defined in subclasses.
Public method get_targeted_logits should be called instead.

interpreto/model_wrapping/inference_wrapper.py
```diff
@@ -631,31 +642,99 @@ def get_gradients(
...
+                outputs=logits,
+                inputs=inputs_embeds,
+                grad_outputs=grad_outputs,
+                retain_graph=(k != t - 1),
+                create_graph=False,
+            )[0]  # (n, l, d)
+
+            # apply the input_x_gradient trick if required
+            if input_x_gradient:
+                target_wise_grads = target_wise_grads * inputs_embeds
```
thomas-mullor, line 675: It would use more space but could be a bit more efficient to do these operations after the for loop ?
Also, you can use in-place multiplication here

tests/attributions/methods/test_attribution_methods_NLP.py
```diff
@@ -63,7 +63,7 @@
         "noise_std": 0.001,
     },
     Saliency: {},
-    IntegratedGradients: {"n_perturbations": 3, "baseline": 0.0},
```
thomas-mullor, line 66: baseline at zero should be tested too

interpreto/model_wrapping/inference_wrapper.py
AntoninPoche: Well, it would, if we did not aggregate over the d dimension in the loop.

I think aggregating in the loop saves a lot of memory and is wise to keep.

But I am open to the debate.

interpreto/attributions/perturbations/linear_interpolation_perturbation.py
AntoninPoche: In theory, yes, but the second condition `baseline == 0` when baseline is a tensor raises the following error:

`RuntimeError: Boolean value of Tensor with more than one value is ambiguous`


Hence, the `isinstance` is a necessary safeguard.

Issue #101, opened 2025-10-07
Casting error with quantized models


ClementBM-CLB:
## 🐛 Bug Report

When applying attribution with a quantized model, various casting errors occur.

## 🔬 How To Reproduce

1. Load a model in quantized mode
2. Run an explaination with any attribution method
3. An error is raised from a casting error, generally because two matrices have different types (e.g. float and c10:Half).

### Code sample

In `test_attribution_methods_NLP.py` change the following:
```python
quantization_config = BitsAndBytesConfig(
    load_in_4bit=True,
    bnb_4bit_use_double_quant=True,
    bnb_4bit_quant_type="nf4",
    bnb_4bit_compute_dtype=torch.bfloat16,
)
model = model_loader.from_pretrained(model_name, quantization_config=quantization_config).to(DEVICE)
```

And then:

```python
if __name__ == "__main__":
    test_attribution_methods_with_text_short_quant(
        model_name="facebook/opt-125m",
        attribution_explainer=IntegratedGradients,
    )
```

### Environment

* OS: Linux, Ubuntu 20.04.6 LTS Focal
* Python version: 3.116
* Interpreto version: 0.4.5

## Expected behavior

Half precision tensor should be casted to full precision?

PR #104, opened 2025-10-08
Solving dtype issue

thomas-mullor:
## Description

Add better dtype management in the inference wrapper
"to" method of the wrapper now follows torch-like signature and more general behaviour :
`my_inference_wrapper.to(device=some_device, dtype=some_dtype)`

If embeddings are passed directly to the inference wrapper, they will be casted to the adapted type (allows dealing with quantized models)

Also add light dtype management for aggregators

## Related Issue

Solves #101

## Type of Change

## Checklist

### DIFF
--- a/interpreto/model_wrapping/inference_wrapper.py
+++ b/interpreto/model_wrapping/inference_wrapper.py
@@ -33,7 +33,7 @@ processing. The class is designed to be subclassed for specific model types and 
 from __future__ import annotations
 
 import warnings
-from abc import ABC, abstractmethod
+from abc import ABC, ABCMeta, abstractmethod
 from collections.abc import Callable, Generator, Iterable, MutableMapping
 from enum import Enum
 from functools import singledispatchmethod
@@ -187,16 +187,16 @@ class InferenceWrapper(ABC):
         Args:
             device (torch.device): wanted device (e.g., "cpu" or "cuda").
         """
-        self.model.to(device)  # type: ignore
+        self.model.to(device)
 
-    def to(self, device: torch.device):
+    def to(self, device: torch.device, dtype: torch.dtype | None = None):
         """
         Move the model to the specified device.
 
         Args:
             device (torch.device): The device to which the model should be moved.
         """
-        self.device = device
+        self.model.to(device=device, dtype=dtype)
 
     def cpu(self):
         """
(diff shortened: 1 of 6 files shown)


interpreto/model_wrapping/inference_wrapper.py
```diff
@@ -210,6 +211,14 @@ def cuda(self):
         """
         self.device = torch.device("cuda")
 
+    @property
+    def dtype(self):
+        return self.model.dtype
+
+    @dtype.setter
+    def dtype(self, dtype: torch.dtype):
+        self.model.to(dtype=dtype)
```
AntoninPoche, line 228: Is it normal to be able to set the dtype of the model?

tests/model_wrapping/test_inference_wrapper.py
```diff
@@ -0,0 +1,46 @@
...
+    # Prepare dummy input
+    input_ids = torch.tensor([[1, 2, 3, 4]])
+
+    # Perform inference
+    outputs = wrapped_model.get_logits({"input_ids": input_ids, "attention_mask": torch.ones_like(input_ids)})
+
+    # Assertions
+    assert outputs is not None
+    assert isinstance(outputs, torch.Tensor)
+    assert outputs.shape[0] == input_ids.shape[0]  # Ensure batch size matches
```
AntoninPoche: I would add a test where you do a forward from `input_embeds` in `torch.float32`.

In our code, this might happen with gradient-based methods.

Because the transformation from `input_ids` to `input_embeds` is managed by the model otherwise.

tests/model_wrapping/test_inference_wrapper.py
AntoninPoche, line 1: Should you not also include a test with each attribution method? To ensure that the whole perturbation-inference-aggregation pipeline is robust to quantized models?

Also, at least a classification and a generation model.

A WORD granularity to be sure granularity manipulations also work with this. (Not the TOKEN default).

interpreto/model_wrapping/inference_wrapper.py
thomas-mullor: It is ok in Pytorch : https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.to

PR #106, opened 2025-10-10
Introduce Insertion and Deletion metrics for attribution methods

cofri:
## Description

This PR introduces the Insertion and Deletion metrics for attribution methods. It includes the following new features:
- new `InsertionPerturbator` and `DeletionPerturbator` that perturbs the inputs based on the order of importance of the tokens. The importance is given by the attribution scores. The Perturbators return masks where elements are iteratively inserted/removed based on the importance. Commit fc488e4
- the new metrics: `Insertion` and `Deletion`. They are built as an AttributionExplainer with custom perturbator and aggregator (defined above). These new classes return an average AUC score and the raw scores for each sequence with the successive perturbations. The raw scores are useful to plot the curve on which the AUC score is computed. Commit f469cfb
- Unit tests are added for these new classes. Commits ef5a787 and 86fac5f

For a seamless integration, two changes are proposed:
- the `AttributionOutput` was extended to store the "model_inputs_to_explain", the targets, the granularity and the inference mode. Commit 5fc08b6
- the base `Perturbator` now accepts kwargs that are transferred to the `get_mask` method. This is required to give the attribution outputs to `get_mask` in order to compute the mask based on the importance of the attributions. Commit 1fd264b

Finally, some bugs are fixed:
- The Generation task now returns targets as 1-D tensor of shape `(t,)` instead of 2-D tensors of shape `(1, t)`. Commit b2e57d4
- Avoid circular imports for `SingleAttribution` type by moving the definition to `typing.py`. Commit 3e0e06e

> [!WARNING]  
> For now, metrics do no support Generation task because of missing aggregations along the target tokens of a same granular element (e.g. aggregation on the n tokens of a word). Support of Generation task will be added when aggregation will be fixed.

## Type of Change

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "interpreto"
-version = "0.4.7"
+version = "0.4.8"
 description = "Interpretability toolbox for LLMs"
 readme = "README.md"
 requires-python = ">=3.10"


interpreto/attributions/base.py
```diff
@@ -119,25 +132,37 @@ def __init__(
...
             f"attributions={self.attributions}, "
             f"elements={self.elements}, "
+            f"model_inputs_to_explain={self.model_inputs_to_explain}, "
+            f"targets={self.targets}, "
             f"model_task='{self.model_task}', "
-            f"classes={self.classes})"
+            f"classes={self.classes}), "
+            f"granularity={self.granularity}, "
+            f"inference_mode={self.inference_mode.__name__}"
         )
```
AntoninPoche, line 162: If PR #105 passes before this one, then the `AttributionOutput` object will have to be adapted.

Notably, the `__repr__` and `__str__` methods have been removed.

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -0,0 +1,282 @@
...
+            device=device,
+            perturbator=self._perturbator_class(
+                tokenizer=self.tokenizer,
+                granularity=granularity,
+                n_perturbations=n_perturbations,
+                max_percentage_perturbed=max_percentage_perturbed,
+                replace_token_id=replace_token_id,
+            ),
+            granularity=granularity,
+            inference_mode=InferenceModes.SOFTMAX,
```
AntoninPoche: You force the `inference_mode`, is there a deep reason?

It is the most adapted, and people should use it.

But did we discuss that no other choice makes sense?

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -0,0 +1,282 @@
...
+    deleted from the input text based on their importance scores. The perturbator class is responsible for
+    handling the specific perturbation logic (insertion or deletion).
+    """
+
+    def __init__(
+        self,
+        model: Any,
+        tokenizer: PreTrainedTokenizer,
+        batch_size: int = 4,
+        granularity: Granularity = Granularity.TOKEN,
```
AntoninPoche: The `granularity` should not be an argument.

- It can be extracted from the `AttributionOutput` in the `evaluate` method.
- It is possible to set the `InsertionDeletionPerturbator` at the beginning of the  `evaluate` method.
- It should still be verified that all `AttributionOutput` have the same `granularity`
- Ultimately, it simplifies the API.

interpreto/attributions/perturbations/insertion_deletion_perturbation.py
```diff
@@ -0,0 +1,151 @@
...
+        """Returns the baseline mask for the perturbator, i.e. a mask with all zeros for
+        deletion and a mask with all ones for insertion.
+
+        Returns:
+            torch.Tensor: Tensor of shape ``(n_perturbations, mask_dim)``
+        """
+        raise NotImplementedError()
+
+    @jaxtyped(typechecker=beartype)
+    def get_mask(
```
AntoninPoche: `    def get_mask(  # type: ignore[override]`

interpreto/attributions/perturbations/insertion_deletion_perturbation.py
```diff
@@ -0,0 +1,151 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Pertubation for the insertion and deletion metric.
```
AntoninPoche: Perturbation for the insertion and deletion metric.

spelling mistake in `Pertubation` -> `Perturbation`

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -0,0 +1,282 @@
...
+        grans = [a.granularity for a in attributions_outputs]
+        if not all(g == self.granularity for g in grans):
+            raise ValueError(
+                f"All attributions must have the same granularity as the insertion/deletion metric. "
+                f"Expected {self.granularity} for all attributions."
+                "If you need to use attributions with different granularities, please raise an issue on GitHub."
+            )
+
+        # Perturb inputs
+        def pertubation_generator(attributions_outputs):
```
AntoninPoche: `        def pertubation_generator(attributions_outputs):`

Same spelling mistake.

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -0,0 +1,282 @@
...
+        target_generator: Iterable[torch.Tensor | None]
+        attrib_idx_gen, pert_generator, target_generator = split_iterator(pertubation_generator(attributions_outputs))
+
+        # Compute the score on perturbed inputs
+        scores: Iterable[torch.Tensor] = self.get_scores(pert_generator, target_generator)
+
+        # Group scores by attribution (i.e. all targets of the same sentence are grouped together)
+        grouped_scores: list[torch.Tensor] = []
+        for _, score in itertools.groupby(zip(attrib_idx_gen, scores, strict=True), key=lambda x: x[0]):
+            grouped_scores.append(torch.stack([s.squeeze(dim=-1) for _, s in score], dim=0))
```
AntoninPoche: Would it make sense to squeeze the stacked tensor and not inside the list comprehension?

If not, just ignore.

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -0,0 +1,282 @@
...
+
+        # Compute AUC using trapezoidal rule
+        scores_np: list[np.ndarray] = [s.cpu().numpy() for s in grouped_scores]
+        aucs = []
+        for s in scores_np:
+            auc = ((s.sum(axis=1) - 0.5 * (s[:, 0] + s[:, -1])) / (len(s.T) - 1)).mean()  # Average AUC across targets
+            aucs.append(auc)
+        auc = float(np.mean(aucs))  # Average AUC across sequences
+
+        return auc, scores_np
```
AntoninPoche: Why are you using numpy here?

The conversion does not seem necessary to me.

interpreto/attributions/base.py
AntoninPoche, line 1: I don't think this commit has its place here. (fix: shape of targets in Generation task is squeezed)

The actual `targets` expected shape by `GenerationInferenceWrapper` is completely underspecified.

Such an update should be accompanied by jaxtyping of the inference wrapper and updates of the docstrings.

But it passes the tests, harmonizes things, and is useful to you. I am not against it passing.

tests/attributions/perturbations/test_insertion_deletion_perturbator.py
```diff
@@ -0,0 +1,137 @@
...
+        DeletionPerturbator(n_perturbations=0)
+
+
+def test_max_percentage_perturbed_zero():
+    """Asserts that the max_percentage_perturbed parameter works correctly when set to zero."""
+    # Max 0% -> still generates at least 2 perturbations
+    perturbator = InsertionPerturbator(n_perturbations=10, max_percentage_perturbed=0.0)
+    attributions = torch.rand(10)
+    mask = perturbator.get_mask(mask_dim=10, attributions=attributions)
+    assert mask.shape == (2, 10), "Should still generate at least 2 perturbations"
```
AntoninPoche: When there are only two elements, no perturbation, and everything perturbed, the metric does not take into account the explanation.

Therefore, it does not make sense.

I would raise a RuntimeTime error, and suggest to increase the number of perturbations or the percentage of perturbed elements.

Maybe a ceil on the linspace could prevent some more corner cases. (Like only one word in the input but max percentage perturbed under 0.5).

tests/attributions/metrics/test_insertion_deletion.py
```diff
@@ -0,0 +1,166 @@
...
+        "task": "classification",
+        "metric_class": Deletion,
+        "granularity": Granularity.WORD,
+        "n_perturbations": 5,
+        "expected_results": {
+            "perturbator_type": DeletionPerturbator,
+            "expected_auc": 0.505140,
+            "scores_shape": [(1, 6), (1, 4), (1, 3)],
+        },
+    },
```
AntoninPoche, line 77: How did you obtain the `expected_auc`?

If it was thanks to a previous run of the method, it does not really test.

It makes sense for non-regression tests. In this case, I would suggest phrasing it that way and making it clear through the assertion message.

In fact, in general, assertion messages in the tests are a good habit to quickly understand what we broke.

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class Insertion(InsertionDeletionBase):
+    """
+    The insertion metric measures the quality of an attribution method by evaluating how the prediction score of a
```
AntoninPoche, line 178: Put in evidence at the top: **Lower is better**

I expect many people to check the documentation just for this.

Same for `Deletion`, but with **Higher is better**

interpreto/attributions/base.py
cofri: Yes, you're right. Depending on the order of the merges, I will update it (or not)

interpreto/attributions/metrics/insertion_deletion.py
cofri: I suppose that this is the only possible mode that makes sense for the AUROC (between 0 and 1). Pure logits can be either positive or negative and log_softmax is always negative. Computing AUROC on these quantities could be difficult to analyze. But if needed, we could let the user choose the inference mode he/she wants.

interpreto/attributions/metrics/insertion_deletion.py
cofri: Yes, good idea to move the instantiation of the Perturbator in the `evaluate()` method

interpreto/attributions/metrics/insertion_deletion.py
AntoninPoche: I agree with you, but maybe a comment justifying this can be helpful for people stumbling upon this.

It could even be precised in the docstring: `scores` will be between 0 and 1 because we look at the evolution of softmax outputs.

interpreto/attributions/perturbations/insertion_deletion_perturbation.py
cofri: Why is this `type: ignore` needed here but not in the other perturbators?

interpreto/attributions/metrics/insertion_deletion.py
AntoninPoche: Without moving the instantiation, you can just update the `self.perturbator.granularity` in the `evaluate()`.

interpreto/attributions/perturbations/insertion_deletion_perturbation.py
AntoninPoche: Here, you override the `get_mask` method in an incompatible manner (if I believe my type checker).

The **kwargs parameter becomes `attributions` or something similar.

The type checker do not like this, (if you use `pyright` with the `standard` type checking it should be underlined in red for you too).

Maybe other `get_mask` also need it since the `**kwargs` parameter introduction.

tests/attributions/metrics/test_insertion_deletion.py
cofri: It's completely for non-regression and not for real value comparison. I'll make it clearer.

interpreto/attributions/perturbations/insertion_deletion_perturbation.py
cofri: Pylance is kinder with me than Pyright :)

PR #107, opened 2025-10-15
Compute gradient by perturbation

ClementBM-CLB:
## Description

* Trying to compute the gradient by perturbation instead of all at once.
* Increases computation time but reduces memory footprint.

## Related Issue

#96 

## Type of Change

## Checklist


AntoninPoche:
Hello @ClementBM-CLB,

Well, this is not precisely the case. In fact, the batching is included for the forward, but not the backward.

Solving this requires some effort, and we are working on it.

Nonetheless, for now, gradient manipulation is done sample by sample. This means that the number of elements in the backward corresponds to the number of perturbations/interpolations requested. Therefore, to reduce the memory impact for gradients, the best is to reduce the number of perturbations required.

We know it is not optimal, but in the short term, this is the best fix we can propose.

PR #113, opened 2025-11-17
Fix generation part for insertion/deletion metrics

fanny-jourdan:
## Description

Minor modification of insertion/deletion to make the metrics work for the generative case and addition of normalization in the generative case for better interpretation of the results (because softmax on the entire vocabulary made the scores incomprehensible). Addition of additional tests for generative cases.

## Type of Change

## Checklist

### DIFF
--- a/interpreto/attributions/metrics/insertion_deletion.py
+++ b/interpreto/attributions/metrics/insertion_deletion.py
@@ -122,16 +122,6 @@ class InsertionDeletionBase(MultitaskExplainerMixin, AttributionExplainer):
             "Use the `evaluate` method to evaluate the deletion metric."
         )
 
-    def _normalize_curve(self, s: torch.Tensor):
-        """
-        Normalize a score curve using its endpoints, depending on the metric type.
-        This keeps the relative shape of the curve and only rescales it
-        to the [0, 1] range based on its endpoints.
-        The normalization is used only for generative models, because the score is too
-        low due to the large vocabulary size when using softmax probabilities.
-        """
-        raise NotImplementedError()
-
     def evaluate(
         self, attributions_outputs: Iterable[AttributionOutput]
     ) -> tuple[float, list[Float[torch.Tensor, "t p"]]]:


.pre-commit-config.yaml
```diff
@@ -22,7 +22,7 @@ repos:
         exclude: LICENSE
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.12
+    rev: v0.14.4
```
cofri: Just in case, latest version is now `0.14.5`

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -43,7 +43,7 @@
 from interpreto.attributions.base import (
     AttributionExplainer,
     AttributionOutput,
-    GenerationAttributionExplainer,
+    # GenerationAttributionExplainer,
```
cofri: This import can be removed

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -86,8 +86,8 @@ def __init__(
                 perturbed (i.e. only the 50% most important). This is useful to avoid perturbing too many elements with
                 low scores in long sequences.
         """
-        if isinstance(self, GenerationAttributionExplainer):
-            raise NotImplementedError("Insertion and Deletion metrics are not implemented yet for generation tasks.")
+        # if isinstance(self, GenerationAttributionExplainer):
+        #    raise NotImplementedError("Insertion and Deletion metrics are not implemented yet for generation tasks.")
```
cofri: Same as my previous comment, these two lines can be removed.

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -138,6 +181,7 @@ def evaluate(
                 - the average AUC score across all attributions
                 - a list of scores for each sequence, where each element is a torch tensor of scores.
         """
+        attributions_outputs = list(attributions_outputs)
```
cofri: Is there a reason to convert this variable as a list?

tests/attributions/metrics/test_insertion_deletion.py
```diff
@@ -165,47 +187,5 @@ def test_non_regression_insertion_deletion(sentences, metric_test_case):
         "The computed AUC does not match the expected value. This test is a regression test, the expected value is "
         "computed from a previous implementation of the metric and ensures no changes in the result."
     )
+    print("metric_scores shapes:", [s.shape for s in metric_scores])
```
cofri: I think you forgot to remove this "print" line

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -124,6 +124,49 @@ def explain(self, *args, **kwargs):  # type: ignore (incoherent override)
...
+        if s.numel() == 0:
+            return s
+
+        s0 = s[0]
+        sP = s[-1]
+        eps = 1e-12
+        name = type(self).__name__
+
+        # Deletion: map start to 1, end to 0 using the endpoints
+        if "Deletion" in name:
```
cofri: Instead of having a `if` condition in the base class, I think it's better to create child methods `_normalize_curve` for `Insertion` and `Deletion`. There is no need then for `raise ValueError(f"Unknown metric type: {name}")`

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -124,6 +124,49 @@ def explain(self, *args, **kwargs):  # type: ignore (incoherent override)
...
         )
 
+    def _normalize_curve(self, s: torch.Tensor) -> torch.Tensor:
+        """
+        Normalize a score curve using its endpoints, depending on the metric type.
+
+        For Deletion:
+            - s_norm[0] ≈ 1 and s_norm[-1] ≈ 0 when the score decreases as expected.
+        For Insertion:
+            - s_norm[0] ≈ 0 and s_norm[-1] ≈ 1 when the score increases as expected.
```
cofri: Just a naïve question: normalization will change the AUROC, is it ok?
What if for instance the curve is negative but the normalization makes it between 0 and 1, does it still make sense?

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -124,6 +124,49 @@ def explain(self, *args, **kwargs):  # type: ignore (incoherent override)
...
+        s0 = s[0]
+        sP = s[-1]
+        eps = 1e-12
+        name = type(self).__name__
+
+        # Deletion: map start to 1, end to 0 using the endpoints
+        if "Deletion" in name:
+            denom = s0 - sP
+            if denom.abs() < eps:
+                # Almost flat curve: neutral constant
```
cofri: It might be dangerous to make the curve flat. What if the first and last element of `s` are equal but there are a lot of variations in between?

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -149,9 +193,16 @@ def evaluate(
         # Perturb inputs
         def perturbation_generator(attributions_outputs):
             for i, a in enumerate(attributions_outputs):
-                for target, attrib in zip(a.targets, a.attributions, strict=True):
-                    pert, mask = self.perturbator.perturb(a.model_inputs_to_explain, attributions=attrib)
-                    yield i, pert, target.to(self.device)
+                is_generation = hasattr(a, "model_task") and "GENERATION" in str(a.model_task)
```
cofri: Maybe using `isinstance(self, GenerationAttributionExplainer)` is cleaner to detect if a model is a generation one

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -149,9 +193,16 @@ def evaluate(
...
+                if is_generation:
+                    for k, attrib in enumerate(a.attributions):
+                        attrib_causal = torch.nan_to_num(attrib, nan=-float("inf"))
```
cofri: This seems to be the role of the perturbator. It could be done directly in the `InsertionDeletionPerturbator`

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -163,8 +214,23 @@ def perturbation_generator(attributions_outputs):
 
         # Group scores by attribution (i.e. all targets of the same sentence are grouped together)
         grouped_scores: list[torch.Tensor] = []
-        for _, score in itertools.groupby(zip(attrib_idx_gen, scores, strict=True), key=lambda x: x[0]):
-            grouped_scores.append(torch.stack(tuple(s for _, s in score), dim=0).squeeze(dim=-1))
+        for idx, score in itertools.groupby(zip(attrib_idx_gen, scores, strict=True), key=lambda x: x[0]):
+            # Determine if this group corresponds to a generation model
+            is_generation = hasattr(attributions_outputs[idx], "model_task") and "GENERATION" in str(
```
cofri: if `is_generation` is used in two different places, you could create only one single `is_generation` variable at the beginning of the function (or elsewhere).

cofri (commented):
Thanks for taking over this generation part of the metric!

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -124,6 +124,49 @@ def explain(self, *args, **kwargs):  # type: ignore (incoherent override)
...
+            denom = sP - s0
+            if denom.abs() < eps:
+                # Almost flat curve: neutral constant
+                return torch.full_like(s, 0.5)
+            s_norm = (s - s0) / (denom + eps)
+
+        else:  # Should not happen
+            raise ValueError(f"Unknown metric type: {name}")
+
+        return s_norm.clamp(0.0, 1.0)
```
cofri: Clamping is also unsafe if the scores between `s0` and `sP` exceeds 0 and 1 (except if we really don't care about these variations)

interpreto/attributions/metrics/insertion_deletion.py
fanny-jourdan: Thank you for pointing that out (I forgot to remove it; it was a debugging aid).

PR #116, opened 2026-01-08
Attribution faithfulness metrics for generation

AntoninPoche:
## Description

#106 introduced `Insertion` and `Deletion` for classification. This pull request extends it to generation.

**This PR should be reviewed commit by commit.**

A few choices were made, which distance the initial idea from blending everything with the existing API. But the code is easier to read. Here is a detail of the choices and justifications:

### Architecture choice

As these metrics are perturbation-based, it was initially decided to inherit from `AttributionExplainer` and `IdsPerturbation` for the different elements of the metrics. However, at most 30 lines (including the multi-task mixin) were helpful in the metrics. While inheriting from `AttributionExplainer` broke most of the rules regarding inheritance. Therefore, it was decided to remove the inheritance. **I am confident this is the correct choice**

The inheritance for the perturbation was also removed. It was either this or the addition of `**kwargs` and a slight behavior change depending on them. However, here, the number of lines copied is much larger. Therefore, I am less confident in this choice. **I welcome your opinion.**

Finally, the aggregation, in #106, included a few lines in the `evaluate function`. However, with the generation of new lines, these lines got extended, so I made it a function. I did not make a class or make it inherit from `Aggregator`. A class with only static methods is useless.

### Details the behavior

Supposing the reader knows how the Insertion and Deletion metrics work. This is to describe the tricks:

Base class for the insertion and deletion metrics for generation models.

The difference with generation is what is explained, classes for classification, and generated tokens for generation. As the perturbation depends on the explanation, different subjects of the explanations lead to different structures of perturbations.

In the case of generation, for a given sample, for each of the $t$ predicted granular elements, the attributions are of length $g$. $g$ corresponds to the number of granular elements (tokens, words, ect. depending on the granularity) preceding the predicted granular element. In practice, the explanations are stored in a matrix of shape $(t _ {max}, g _ {max})$ with NaN values, but this has no importance here.

In any case, for each sample, there are $t$ explanations. Thus, the insertion/deletion metric should be computed for each one of them. Hence, for generation, we iterate first on the samples, then on the explanations. The perturbations are computed for each sample-explanation pair.

The problem is that for a given attribution, thus a given sample-explanation pair, the insertion/deletion perturbation should only be done on the inputs used to predict the explained predicted granular element. We could perform perturbation on all elements, but in the worst case, it results in twice as many inferences and provides less precise values.

Hence, we iteratively include one more granular element at a time and set this element as the target. To understand the relationship between granular elements and tokens, we utilize granularity indices.

Example: (keep in mind that for generation, the targets are included in the input to explain)
initial input: "A BC", initial target "DEF GHIJ KLMNOP"
pert 1 input: "A BC DEF", pert 1 target "DEF"
pert 2 input: "A BC DEF GHIJ", pert 2 target "GHIJ"
pert 3 input: "A BC DEF GHIJ KLMNOP", pert 3 target "KLMNOP"

### The tests

Naturally, this PR comes with tests. The classification tests were extended to generation and completed with many assertions.

Moreover, a qualitative test was introduced with a BERT model trained on IMDB. The qualitative test checks that attributions with orders are either good, random, or bad and ranks them accordingly using the metrics.

## Type of Change

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "interpreto"
-version = "0.4.11"
+version = "0.4.12"
 description = "Interpretability toolbox for LLMs"
 readme = "README.md"
 requires-python = ">=3.10"


interpreto/attributions/perturbations/insertion_deletion_perturbation.py
```diff
@@ -53,33 +53,68 @@ class InsertionDeletionPerturbator(IdsPerturbator):
...
     correspond to the first target, the next `p` rows to the second target, and so on.
 
-    Args:
+    Attributes:
         tokenizer (PreTrainedTokenizer | None): Hugging Face tokenizer associated with the model.
-        granularity (Granularity): Level at which deletion should be applied.
+        granularity (Granularity): Level at which deletion should be applied. Set after initialization.
         n_perturbations (int): Number of perturbations to generate.
         max_percentage_perturbed (float): Maximum percentage of tokens in the sequence to be perturbed.
         replace_token_id (int): Token used to replace deleted elements.
```
cofri, line 61: Missing ­­`device` in docstring

interpreto/attributions/perturbations/insertion_deletion_perturbation.py
```diff
@@ -53,33 +53,68 @@ class InsertionDeletionPerturbator(IdsPerturbator):
...
+        Get the device of the perturbator
+        """
+        return self._device
+
+    @device.setter
+    def device(self, device: torch.device):
+        """
+        Set the device of the perturbator
+        """
+        self._device = device
```
cofri: This getter and setter do not bring any processing that justifies the need for a getter and a setter. I suggest to directly use a class member `self.device`, and remove `self._device`.

interpreto/commons/granularity.py
cofri: `get_indices` is also called in `granularity_score_aggregation`. Is it interesting to also add the precomputed indices as argument in this method?

interpreto/commons/granularity.py
```diff
@@ -218,7 +218,7 @@ def get_indices(
 
                 n_inputs = inputs["input_ids"].shape[0]  # type: ignore
 
-                if tokenizer.is_fast:
+                if tokenizer.is_fast and isinstance(inputs, BatchEncoding):
```
cofri, line 221: Out of curiosity, why did you add the if condition `BatchEncoding` here? Is it required?

interpreto/attributions/metrics/insertion_deletion.py
```diff
@@ -86,115 +96,415 @@ def __init__(
...
+    In most cases, a single class is explained for each sample, so the number of targets is 1.
+
+    In any case, for each sample, there is $t$ explanations.
+    Thus the insertion/deletion metric should be computed for each one of them.
+    Hence, for classification, we iterate first on the samples, then on the explanations.
+    The perturbations are computed for each sample-explanation pair.
+    """
+
+    _associated_inference_wrapper = ClassificationInferenceWrapper
+    inference_wrapper: ClassificationInferenceWrapper
```
cofri, line 359: Why is this line required?

interpreto/commons/granularity.py
AntoninPoche: Yes, it would be pertinent.

interpreto/commons/granularity.py
AntoninPoche: There are two possible paths here:
- either no word_ids, the token ids are converted back to text, and we detect the words in the text.
- or using word_ids, which is a parameter of the `BatchEncoding` object if the latter is the output of a `FastPreTrainedTokenizer`'s `__call__()` method.

In interpreto, sometimes you will find written `TensorMapping`, which is our internal type for a dictionary of `torch.Tensor`. In practice, `BatchEncoding` acts as a `TensorMapping` (true only for `return_tensor="pt"`). But it also has few attributes and methods, such as `word_ids` (only present in ).

Therefore, to go in the second case, we need both conditions to be true.

But we could also check with `hasattr(inputs, "word_ids")`.

interpreto/attributions/metrics/insertion_deletion.py
AntoninPoche: For typing purposes. In the base class, we have `inference_wrapper: InferenceWrapper`, which is underspecified now.

I needed to do it in `AttributionExplainer` for `pyright` to consider the code correct; thus, I copied it, but it may not change anything.

PR #126, opened 2026-02-10
Fix links of tutorials in readme and doc

gfouilhe:
There was an extra "s" which broke the links

## Type of Change


AntoninPoche:
Nice catch! Thank you @gfouilhe !

PR #133, opened 2026-03-09
Fix bug for word and sentence granularity

fanny-jourdan:
## Description

- Fixed a bug in word and sentence granularity for certain models
- Fixed a bug in the visualizations for attribution.
- Remove the possibility to use the attribution method with target=None for Generation Task
- Clean some tests 

## Type of Change

## Checklist


interpreto/attributions/base.py
```diff
@@ -400,9 +407,6 @@ def explain(
             for score, mask in zip(scores, mask_generator, strict=True)
         )
 
-        # Aggregate the score with respect to the granularity level
-        # - Aggregate over the inputs for gradient-based methods: (t, l) -> (t, lg)
-        # - Aggregate over the targets if the model is a generation model: (t, l) -> (tg, l)
```
AntoninPoche, line 405: Why remove the comments?

I think that it is useful to understand what the following few lines do without opening the `granularity.py` file. Which is, IMO, far from trivial.

interpreto/commons/granularity.py
```diff
@@ -308,6 +309,18 @@ def __token_get_indices(tokens_ids: torch.Tensor, special_ids: list[int]) -> lis
         """Indices for :pyattr:`TOKEN` – skip special tokens."""
         return [[i] for i, tok_id in enumerate(tokens_ids) if tok_id not in special_ids]
 
+    @staticmethod
+    def __word_ids_are_usable(tokenizer: PreTrainedTokenizer | PreTrainedTokenizerFast, inputs: BatchEncoding) -> bool:
+        """Return True when we have a fast-tokenizer and word ids provide meaningful word grouping."""
+        if not tokenizer.is_fast:
+            print("Tokenizer is not fast, cannot use word_ids for WORD granularity.")
+            return False
+        word_ids = inputs.word_ids(0)
```
AntoninPoche, line 318: Are we sure that fast tokenizers produce word_ids? What I mean by that, is that, is there no risk of `AttributeError inputs has no attribute word_ids`?

interpreto/attributions/base.py
```diff
@@ -616,6 +620,27 @@ def process_inputs_to_explain_and_targets(
...
+        ]
+    else:
+        raise TypeError(f"Decoded target text must be a string or an iterable of strings, got {type(target_text)}.")
+    normalized_target_ids = tokenizer(
+        normalized_target_text,
+        return_tensors="pt",
+        truncation=True,
+        add_special_tokens=False,
+    )["input_ids"].squeeze(dim=0)
+    return normalized_target_ids
```
AntoninPoche, line 646: In this function, you force the decode and encode. But if the beginning space is already present, then the provided `target` tensor received is correct. Then we should return `target` directly, not `normalized_target_ids`.

Are there any other necessary steps other than adding this space?

interpreto/attributions/base.py
AntoninPoche, line 695: In this function, you call `normalize_target_ids_with_leading_space` function, each time you do so, it seems that the target tensor corresponds to a single input. Therefore, I imagine that you do not need the case where the decoded `target` into `target_text` is a list.

If I am correct, then you might be able to simplify the `normalize_target_ids_with_leading_space` function by removing the list case.

AntoninPoche (approved):
Nice Work @fanny-jourdan !!

I made a few suggestions that I am not sure make sense, as I do not know the exact objects that are manipulated.

interpreto/attributions/base.py
fanny-jourdan: Oh yeah, that's a mistake! Actually, while I was writing the code, I changed a bunch of things that I had commented out to keep track of my own changes, and when I deleted the commented-out lines, I must have accidentally deleted that one too!

interpreto/commons/granularity.py
fanny-jourdan: Yeah, that's usually the case! Even in tokenizers like Mistral, which have a “word id” of NaN, there's still have the option. Of course, we can't rule out the possibility that other, lesser-known tokenizers exist, but for now, I feel like I've covered

interpreto/attributions/base.py
fanny-jourdan: Unfortunately, I have to do this to make sure the space is actually there... I checked, but if I don't do this and we pass the tokenized target directly, it concatenates the last word of the input with the first word of the target.
And since I have to decode it to verify that the user didn't enter just anything and check for a space, I have to re-encode it afterward.

interpreto/attributions/base.py
fanny-jourdan: That's what I did. If it's an iterable, I don't use `normalize_target_ids_with_leading_space`. All other cases are independent.

interpreto/attributions/base.py
AntoninPoche: That was not exactly my point. I understand that you need to convert the `target` to `target_text` to check for spaces. But then you have two cases:
- The space is not present. Then modify `target_text`, encode the modified version to `normalized_target_ids`, and return the last element. I totally agree.
- But there is another case, where the space is present in `target_text`, which means that the `target` was correct, and you can return them directly without using the `normalized_target_ids`.

interpreto/attributions/base.py
AntoninPoche: My comment concerns modifications to `normalize_target_ids_with_leading_space`. (I might not have placed it at the correct spot.)

To my understanding, the `target` tensor you pass to the `normalize_target_ids_with_leading_space` function always has the shape `(l,)`. Therefore, once you convert it to `target_text`, it is a `str` and cannot be a `Iterable[str]`.

Which means that the `elif isinstance(target_text, Iterable):` is not necessary and can be removed. Which makes my prior comment with returning `target` directly much easier to implement.

interpreto/attributions/base.py
fanny-jourdan: Ah ok I see  ! I will do this  !

interpreto/attributions/base.py
fanny-jourdan: I can't remember exactly in what context I encountered an error involving an iterable, which is why I added that safeguard. So it might seem pointless, but I didn't put it there for nothing :)

AntoninPoche: Okay, no problem, but you shouldn't bother solving both comments at the same time. (The early exit for the iterable seems to be a pain.)

PR #134, opened 2026-03-10
Extract ngram

camillebrl:
## Description

Add n-gram support for concept interpretation + notebook tuto

## Related Issue

No existing issue: unique words alone can miss meaningful multi-word expressions (e.g. in NOTAMs, "RWY CLSDE" is more informative than "RWY" or "CLSD" separately). This adds extract_ngrams and extends use_unique_words to accept an integer: 0 for default behavior, 1 for unique words, >1 for n-grams up to that size.

## Type of Change

Feature addition


docs/notebooks/examples/classification_ngram_concept_tutorial.ipynb
AntoninPoche, line 1: It is nice that you made a tutorial, but I have several remarks on it:
- I would reduce it to make it less like the others, but focus on the interpretation part. In particular, say in the beginning what this notebook is for and quickly go through the first steps in the "Construct concept-space" first section.
- You should move it to the sub-directory `docs/notebooks/examples/`.
- I would use SemiNMF rather than ICA to have better concepts and better highlight the interpretation method.
- Not mandatory, but you can also increase the number of samples in the concept-space training to obtain better concepts.

You are free to ignore these comments if you want. The interpretations are great, congrats!

I think I will use this for LLM labels in classification.

interpreto/concepts/interpretations/base.py
AntoninPoche, line 1: Seeing how similar the `extract_unique_words` and `extract_ngrams`, you can remove the `extract_unique_words` one.

interpreto/concepts/interpretations/base.py
```diff
@@ -167,6 +167,80 @@ def extract_unique_words(
...
+
+        # extract n-grams of all sizes from 1 to n
+        for size in range(1, n + 1):
+            for i in range(len(processed) - size + 1):
+                ngram = " ".join(processed[i : i + size])
+                ngram_counts[ngram] += 1
+
+    # filter too rare n-grams
+    if count_min_threshold > 1:
+        ngram_counts = Counter({key: count for key, count in ngram_counts.items() if count >= count_min_threshold})
```
AntoninPoche: Small acceleration, by directly adding single words in the first word loop and hashing ngrams as tuples to build them only in the end.

```python
for text in inputs:
    tokens = word_tokenize(text)

    processed = []
    for word in tokens:
        if lemmatize:
            word = lemmatizer.lemmatize(word.lower())  # noqa: PLW2901
        if words_to_ignore is not None and word in words_to_ignore:
            continue

        processed.append(word)
        ngram_counts[(word,)] += 1  # unigram tuple

    for size in range(2, n + 1):  # skips size 1 as covered over
        for i in range(len(processed) - size + 1):
            ngram_counts[tuple(processed[i : i + size])] += 1

ngram_counts = Counter({
    " ".join(k): count  # convert ngram tuples to strings
    for key, count in ngram_counts.items()
    if count >= count_min_threshold
})
```

interpreto/concepts/interpretations/base.py
```diff
@@ -259,7 +333,7 @@ def __init__(
         aggregation_strategy: GranularityAggregationStrategy = GranularityAggregationStrategy.MEAN,
         concept_encoding_batch_size: int = 1024,
         use_vocab: bool = False,
-        use_unique_words: bool = False,
+        use_unique_words: int = 0,
```
AntoninPoche: If you pass them to `int`, I think you break retro compatibility.

It would be best to have `bool | int`.

Then, when you can easily convert it via `self.use_unique_words = int(use_unique_words)`, which converts `False` to `0` and `True` to `1`.

interpreto/concepts/interpretations/base.py
```diff
@@ -283,7 +357,7 @@ def __init__(
         self.aggregation_strategy: GranularityAggregationStrategy = aggregation_strategy
         self.concept_encoding_batch_size: int = concept_encoding_batch_size
         self.use_vocab: bool = use_vocab
-        self.use_unique_words: bool = use_unique_words
+        self.use_unique_words: int = use_unique_words
```
AntoninPoche: same

interpreto/concepts/interpretations/base.py
```diff
@@ -568,13 +642,18 @@ def get_granular_inputs_and_concept_activations(
...
+                # Case 2: use_unique_words >= 1
+                # first list unique words/ngrams from the inputs and compute the activations from them
+                if self.use_unique_words == 1:
+                    granular_inputs: list[str] = extract_unique_words(
+                        inputs=inputs, return_counts=False, **self.unique_words_kwargs
+                    )  # type: ignore  (sure list[str] with return_counts=False)
+                else:
+                    granular_inputs: list[str] = extract_ngrams(
+                        inputs=inputs, n=self.use_unique_words, return_counts=False, **self.unique_words_kwargs
+                    )  # type: ignore  (sure list[str] with return_counts=False)
```
AntoninPoche: No need for the two cases, in my opinion.

interpreto/concepts/interpretations/llm_labels.py
```diff
@@ -176,7 +176,7 @@ def __init__(
         k_examples: int = 30,
         k_context: int = 0,
         use_vocab: bool = False,
-        use_unique_words: bool = False,
+        use_unique_words: int = 0,
```
AntoninPoche: Same

interpreto/concepts/interpretations/topk_inputs.py
```diff
@@ -226,7 +226,7 @@ def __init__(
         concept_encoding_batch_size: int = 1024,
         k: int = 5,
         use_vocab: bool = False,
-        use_unique_words: bool = False,
+        use_unique_words: int = 0,
```
AntoninPoche: Same

AntoninPoche (changes requested):
Great first contributions @camillebrl !!

The concept's interpretations seem much more interpretable than the words alone.

I made a few comments to simplify the code, mainly by removing `extract_unique_words()`.

Also few comments on the notebooks, it depends on how much time you have to work on this, do not over extend on the notebook, I know it can easily get time consuming.

AntoninPoche (approved):

PR #142, opened 2026-04-15
`InferenceWrapper` rewrite and attribution device optimization

### LINKED ISSUE #96: Jacobian calculation taking too much GPU memory with small LLMs

### LINKED ISSUE #108: Refactor InferenceWrapper and subclasses for better readability and gradient handling
**Description:**
The `InferenceWrapper` class and its subclasses could benefit from a refactoring aimed at improving readability and simplifying gradient computation over large input batches. It would also provide batched computation of gradient on large inputs (actually not well managed)

**Goals:**
Refactor the simple `call_model` method to directly handle a wide range of options.
Other externally called methods should delegate to `call_model` directly instead of following the current call chain.

**TODO:**
1. Transform `get_logits` into some `batch_call_model` method that only takes iterables (generators) as input :
Input parameters:
- `input`: (unchanged)
- `target`: Optional. If None, return full input logits; otherwise, return targeted_logits.
- `return_grad`: Boolean, default False. If True, perform backward pass; otherwise, use a no_grad context.
- ...additional options as needed
Responsibilities:
-   Handle context management for different tasks (classification, generation, etc.)
-   Manage input slicing and batching internally instead of in get_logits
The version of `get_logits` that takes mapping as input is simply removed and this kind of usage won't be suported anymore.
Code division : It would be better if the current implementation was split into several specific, clearly named functions and the batching management was more like pseudo code.

2. Update other methods to delegate to `batch_call_model`:
- `get_logits`: simple call to `batch_call_model` with `target=None`
- `get_targeted_logits`: call `batch_call_model` with the appropriate target
- `get_gradients`: call `batch_call_model` with `return_grad=True`
- `get_inputs_to_explain_and_targets` : For now, this method that is specific to generation context could be defined on his side, without dealing with large batching. We assume that users would not want to call generation tasks on a large amount of sentences

3. Code simplification:
Remove input-based dispatching; only support iterable inputs
Centralize sanity checks and input transformations (reshape_inputs, process_inputs, etc.)

**Expected Outcome:**
A cleaner, more maintainable `InferenceWrapper` with a flexible entry point (`call_model`) for all types of input processing, task handling, and gradient computation.

**Collaboration & Feedback:**
This issue is open for proposals. Developers and users are welcome to suggest improvements or related ideas in the comments. Contributions and discussions are encouraged to shape the best possible design.

### LINKED ISSUE #127: Memory management for generation.
## 🐛 Bug Report

With generation, the model's output has the shape (batch, seq_len, vocab_size). The vocab is often super large, ~100k. Hence, for medium seq_len ~50, the batch should be small.

However, the "targeting," as we call it, is done after stacking all elements from perturbations. Therefore, in `_get_logits_from_iterable`, we create a tensor of shape (nb_perturb, seq_len, vocab_size). Where nb_perturb can be greater than the required batch_size and thus cause out-of-memory errors.

To solve this, the targeting should be applied to batches before stacking all perturbed elements.

## 🔬 How To Reproduce

`KernelShap` with a generation LLM

## Additional context

This is closely related to the issue #108

AntoninPoche:
## Description

The overall goal was to simplify the `InferenceWrapper` and control tensor device movement. Some bugs were inherent to the previous code architecture.

Few choices were made to simplify the whole pipeline, with the main idea being to clearly separate roles. This list is detailed later.
- Only the `InferenceWrapper` moves things into the `device`.
- The `InferenceWrapper` does not embed `input_ids`.
- Perturbed inputs for `InferenceWrapper` are already repeated.
- For each sample, the target is a vector of t dimensions.
- We do not support `BatchEncodings` with multiple samples.

Additionally, some minor modifications were applied:
- Better typing in `attributions/base.py`.
- Harmonize and robustify `pad_token_id` setup.

**However, the main modification of this PR is the complete rewrite of the `InferenceWrapper` and children classes.**

### Only the `InferenceWrapper` moves things into the `device`.

Device management was previously split across perturbation, inference wrapper, aggregation, and the base explain. Now, mainly the inference wrapper does.

The `InferenceWrapper` consumes inputs as `TensorMapping` and targets as `Tensor`, which are moved to the device with `_prepare_inputs` when creating a batch. Once processed, these batch inputs are destroyed, so they are confined.

The only thing left on the device is the output (targeted logits or gradients), which the aggregator processes sample by sample. In some cases, the mask needs to be moved to the output device. It is done by the  `cast_input_to_dtype` decorator.

### The `InferenceWrapper` does not embed `input_ids`.

There was a duplicated `embeds` function between `inference_wrapper.py` and `perturbations/base.py`. Everything was moved to the `perturbations/base.py`, in particular, the `EmbeddingsPerturbator`.

Additionally, the `perturb_embeds()` method was simplified, treating only the `inputs_embeds` tensor and not the tensor mapping. Also, the `perturb` has now the role of repeating the other values of the tensor mapping.

### For each sample, the target is a vector of t dimensions.

No matter the perturbation, the target remains the same across all versions of an initial input. `InferenceWrapper` is not user exposed, so it can be rigid. On the other hand, it is the role of the preprocessing functions to harmonize inputs to the expected format. Therefore, this was forced through by typing and by some modifications to the preprocessing.

### We do not support `BatchEncodings` with multiple samples.

When taking multiple input tokens simultaneously, it introduces padding, which conflicts with the new internal padding. Therefore, it was removed. To pass tokenized inputs, users now need to tokenize them individually.

This can be seen as a **breaking change** because we now support less inputs types than before. However, this issue was discovered during the introduction of more comprehensive tests and more precise typing. Therefore, I would rather see it as bug fixing.

### Better typing in `attributions/base.py`.

Everything was typed as `Iterable`, encompassing both `Iterator` and `list`. But these, while being both "iterable", have an important difference. `Iterator` from generators can be iterated only once, while `list` can be iterated indefinitely. Indeed, once consumed, an `Iterator` becomes empty, and its elements are destroyed because nothing references them anymore.

This difference is key in `AttributionExplainer.explain()`, it tells what we can reuse (the model inputs or targets) and what we cannot reuse (perturbed inputs, masks, scores...). The advantage is that we do not need to care about the device of elements in `Iterator`.

### Harmonize and robustify `pad_token_id` setup.

Funnily enough, each model and tokenizer uses a different padding token ID. Similarly, it is stored in different objects. It was thus necessary to make the `setup_token_ids()` function more complex. But it regroups all token ID selection and settings. Therefore, the padding token ID is no longer a wild guess.

### `__call__()`

The only entry point to `InferenceWrapper`. The same function applies to the three behaviors (targets, targeted logits, and gradients). This behavior is mainly managed by `_call_batch()`. In any case, for each input, an output is yielded.

The main role of `__call__()` is to construct `Batch` and call `_call_batch()` on it. Similar to previous implementations, the `InferenceWrapper` fills each `Batch` to `batch_size` to minimize the number of model calls.

However, the algorithm is a bit different. We iterate over samples (which are batches of perturbed inputs), then take the chunk of perturbed inputs needed to fill the `Batch`. If the `Batch` is not full, we take in a new sample. Once the `Batch` is full, we call `_call_batch()`. Note that a sample can be split across several `Batch`

### `_call_batch()`

The `Batch` object links the initial sample chunks to the data sent to the model, and thus to the model's output. A `Batch` becomes a `TensorMapping` through `_prepare_inputs()` (pad and concat).

After the forward pass, the outputs are chunked relative to the corresponding perturbed input chunks. Then, the different behaviors are applied to the chunked logits:
- targets (computing the targets from the inputs for classification)
- targeted logits (computing logits from inputs and extracting the pertinent ones using the targets)
- gradients (gradients of the targeted logits with respect to the inputs)

The behavior is selected through the presence of `targets` and the `gradients` attribute of the `InferenceWrapper`:
- `targets is None`: targets (raise an error for generation)
- `targets is not None and not self.gradients`: targeted logits
- `targets is not None and self.gradients`: gradients

With all the different behaviors, we obtain an output chunk, which is added to a buffer. When we receive the last chunk of an initial sample, we yield the concatenated chunks.

### Task-specific processing

With this rewrite, task-specificity is minimal; it corresponds to the function that targets logits or extracts targets.

For generations, there has also been the addition of a `position_ids` key in `_prepare_inputs()`. It was necessary; otherwise, the padding would change the output. (Which was the case before).

## Related Issue

#96 
#108 
#127 

And several undetected bugs, like the `position_ids` one, the padding from `BatchEncoding`, or the forced `pad_token_id`.

## Type of Change

## Checklist


interpreto/model_wrapping/inference_wrapper.py
```diff
@@ -205,505 +254,356 @@ def to(self, device: torch.device, dtype: torch.dtype | None = None):
...
-                        for inputs_chunk, mask_chunk in zip(inputs_chunks, mask_chunks, strict=False)
-                    ],
+        # for each key we pad the values and aggregate them
+        padded_inputs = {}
+        for key in inputs[0].keys():
+            padded_inputs[key] = pad_sequence(
+                [elem[key] for elem in inputs],
+                batch_first=True,
+                padding_value=self.pad_token_id if key == "input_ids" else 0,
+                padding_side=self.padding_side,
```
AntoninPoche, line 396: Should use `getattr(self.tokenizer, "padding_side", self.padding_side)`. All tokenizers should have a padding_side.

interpreto/model_wrapping/inference_wrapper.py
```diff
@@ -205,505 +254,356 @@ def to(self, device: torch.device, dtype: torch.dtype | None = None):
...
+            last_target = k == (t - 1)
+            # we compute gradients one target at a time to save memory and avoid jacobian computations
+            target_wise_grads: Float[torch.Tensor, f"{c} l d"] = torch.autograd.grad(
+                # sum the targeted logits to get a scalar output for the backward pass, but this does not change the gradients as the targeted logits are independent between each others
+                outputs=targeted_logits_chunk[:, k].sum(),
+                inputs=inputs_embeds,  # type: ignore
+                # we retain the graph for all but the last iteration to avoid unnecessary recomputation of the forward pass
+                retain_graph=not (last_target and last_chunk),
+                # we do not create a new graph as we do not need higher order gradients
+                create_graph=False,
```
AntoninPoche, line 353: It is `False` by default, so this is not necessary.

interpreto/attributions/aggregations/base.py
```diff
@@ -41,10 +41,10 @@ def cast_input_to_dtype(func):
     Ensure mask and results are on the device specified in the aggregator
     """
 
-    def wrapper(self, results: torch.Tensor, mask, *args, **kwargs) -> torch.Tensor:
-        # TODO : eventually add device alignment as well
-        if mask is not None and mask.dtype != self.dtype:
+    def wrapper(self, results: torch.Tensor, mask: torch.Tensor | None, *args, **kwargs) -> torch.Tensor:
+        if mask is not None:
             mask = mask.to(self.dtype)
+            mask = mask.to(results.device)
```
thomas-mullor: 'to' method can handle dtype and device in a single call

interpreto/attributions/perturbations/base.py
```diff
@@ -276,21 +237,22 @@ def perturb(self, model_inputs: TensorMapping) -> tuple[TensorMapping, torch.Ten
...
+            model_inputs,  # type: ignore
+            self.tokenizer,
+        )[0].float()
 
         # compute granularity-wise perturbation mask based on the length of the sequence (granularity-wise)
-        gran_mask: Float[torch.Tensor, "p g"] = self.get_mask(association_matrix.shape[0]).to(self.device)
+        gran_mask: Float[torch.Tensor, "p g"] = self.get_mask(association_matrix.shape[0])
 
         # compute real perturbation mask
         real_mask: Float[torch.Tensor, "p l"] = torch.einsum("pg,gl->pl", gran_mask, association_matrix)
```
thomas-mullor: einsum ab,bc->ac is just a matrix product

interpreto/model_wrapping/classification_inference_wrapper.py
```diff
@@ -21,207 +21,40 @@
...
-                ...                 {"input_ids": torch.tensor([[7, 8, 9], [10, 11, 12]])}]
-                >>> targets = torch.tensor([[1, 2], [3, 4]])
-                >>> target_logits = wrapper.get_targeted_logits(model_inputs, targets)
-                >>> for logits in target_logits:
-                ...     print(logits)
-        """
-        raise NotImplementedError(
-            f"type {type(model_inputs)} not supported for method get_target_logits in class {self.__class__.__name__}"
-        )
+    @property
```
thomas-mullor: Why use a constant property rather than a static attribute ?

interpreto/model_wrapping/inference_wrapper.py
thomas-mullor: "Explicit is better than implicit"

interpreto/model_wrapping/inference_wrapper.py
```diff
@@ -205,505 +254,356 @@ def to(self, device: torch.device, dtype: torch.dtype | None = None):
...
 
     @dtype.setter
     def dtype(self, dtype: torch.dtype):
-        self.model.to(dtype=dtype)
+        self.model.to(dtype=dtype)  # type: ignore
 
-    def embed(self, model_inputs: TensorMapping) -> TensorMapping:
+    @property
+    @abstractmethod
+    def padding_side(self) -> str:
```
thomas-mullor: should be a static attribute

thomas-mullor (approved):
Great work !
Love the new InferenceWrapper.
I left a few minor comments but nothing that would really block the request

PR #153, opened 2026-05-25
Probes

AntoninPoche:
This PR should be reviewed commit by commit.

## Description

Introduce supervised concept-based methods (also called probes or CAVs).

It follows a large part of the existing concept-based methods, which we can now frame as unsupervised. Common points:
- Probes require a `ModelWithSplitPoints`
- Probes should be used through an explainer that takes the `ModelWithSplitPoints` and a concept model.
- The probes are trained via `fit`

Nonetheless, there are a few particularities:
- We only define probe concept models and do not create probe-wise explainers. This means the user must first instantiate the probe before giving it to the explainer.
- The `fit` method takes both the activations and the concept labels.

```python
from interpreto.concepts import LinearRegressionProbe, ProbeExplainer

probe = LinearRegressionProbe()
explainer = ProbeExplainer(model_with_split_points, probe)
explainer.fit(activations, labels)
concepts = explainer.encode_activations(activations)
```

Note that probing is multi-output, allowing us to train several probes simultaneously. This is much more efficient as each probe is independent from the others.

### Types of probes

The above example uses the `LinearRegressionProbe`, but many probes have been implemented. We could distinguish two categories:

#### Linear probes

The final probe computes the concept activations by taking a scalar product between the activations and the weights. Therefore, weights encode directions, and concepts are seen as directions. The linear probes differ in how they find the weights.

#### Centroid-based probes

The final probe computes the concept activations as the biased negative distance toward a centroid. The closer to the centroid, the higher the concept activation. Here, concepts can be seen as clusters (except for the cosine distance one, which corresponds to a direction due to the distance used).

## Type of Change

## Checklist

### DIFF
--- a/docs/api/concepts/probes.md
+++ b/docs/api/concepts/probes.md
@@ -1,0 +1,303 @@
+---
+icon: material/target
+---
+
+# Probes (Supervised)
+
+Probes are **supervised** concept methods: they require labeled data (concept annotations)
+to learn a mapping from activations to concept scores. This contrasts with the unsupervised
+[Concept Spaces](./concept_spaces/base.md) (ICA, NMF, SAEs, …) which discover concepts
+from unlabeled activations alone.
+
+The probe workflow uses the same `ModelWithSplitPoints` activation extraction as unsupervised
+methods, but the `fit` step requires both activations **and** binary concept labels.
+
+## Usage Guide
+
+### Classification Model (SplitSequenceClassification)
+
+The simplest setup for classification models. `SplitSequenceClassification` automatically
+detects the classification head and uses CLS-token activations.
+
+```python
+from interpreto import SplitSequenceClassification
+from interpreto.concepts import ProbeExplainer
+from interpreto.concepts.probes import LinearRegressionProbe
+
+# 1. Wrap your classification model
+model = SplitSequenceClassification("textattack/bert-base-uncased-imdb")
+
+# 2. Extract CLS-token activations — shape (n, d)
+activations = model.get_activations(texts)
+
+# 3. Instantiate probe and explainer
+probe = LinearRegressionProbe()
+explainer = ProbeExplainer(model, concept_model=probe)
+
+# 4. Fit on activations + binary concept labels — labels shape (n, c)
+explainer.fit(activations, labels)
+
+# 5. Score new inputs
+concept_scores = explainer.encode_activations(new_activations)
+```
+
+### Generation Model (ModelWithSplitPoints)
+
+For generation models, you must choose how to aggregate the sequence of token-level
+activations into a fixed-size representation. Two common strategies:
+
+#### Strategy A: Aggregate to one vector per sample
+
+Use `activation_granularity=SAMPLE` to pool all tokens into one activation vector.
+This is appropriate when concepts are global properties of the input (e.g., topic, style).
+
+```python
+from interpreto import ModelWithSplitPoints
+from interpreto.concepts import ProbeExplainer
+from interpreto.concepts.probes import CosineCentroidProbe
+
+model = ModelWithSplitPoints(
+    "gpt2",
+    split_points="transformer.h.6",
+    device_map="cuda",
+)
+
+# Aggregate all tokens into one vector per sample — shape (n, d)
+activations = model.get_activations(
+    texts,
+    activation_granularity=model.activation_granularities.SAMPLE,
+    aggregation_strategy=model.aggregation_strategies.MEAN, # MAX and LAST are also often compared in the literature
+)
+
+probe = CosineCentroidProbe()
+explainer = ProbeExplainer(model, concept_model=probe)
+explainer.fit(activations["transformer.h.6"], labels)
+```
+
+#### Strategy B: Per-token activations (flattened)
+
+Use `activation_granularity=TOKEN` to get one activation per token (special tokens
+removed, then flattened). This is appropriate when concepts are local properties
+(e.g., named-entity type, part-of-speech) and labels are provided per-token.
+
+```python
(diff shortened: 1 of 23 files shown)


docs/api/concepts/probes.md
fanny-jourdan, line 1: You also have to modify docs/index.md

docs/notebooks/classification_probe_tutorial.ipynb
fanny-jourdan, line 1: When you create concepts, wouldn't it be a good idea to create a concept with a set of positive words and a set of negative words (like “awesome,” “favorite,” “good,” etc., versus “worst,” “lame, “bad,” etc.)?

docs/notebooks/generation_probe_tutorial.ipynb
fanny-jourdan, line 1: I'm not sure I understand how the probes work for generative part. Don't we have a predefined concept, like we do for classification?

docs/api/concepts/probes.md
AntoninPoche: Both the readme and the index.md correct?

Saying that now we cover both supervised and unsupervised?

docs/notebooks/classification_probe_tutorial.ipynb
AntoninPoche: We could also add a positive vs negative concept. But is it necessary? By that, do you mean I should update and rerun the tutorial to include it? Does it add much in comparison to the others?

docs/notebooks/generation_probe_tutorial.ipynb
AntoninPoche: The concept we study is truthfulness, as it is studied in the literature. We extract it from a Q&A dataset.

It is at the sample level, word-level annotations are a pain to align.

fanny-jourdan: Ok so you study only one concept ? (it is not a problem i just ask to understand)

docs/notebooks/classification_probe_tutorial.ipynb
fanny-jourdan: I don't think it's “necessary” but I think the tutorial would benefit from it because the example would be very comprehensive (and it's a very natural concept for the task)

docs/notebooks/generation_probe_tutorial.ipynb
AntoninPoche: Yes, it is a single concept

docs/notebooks/classification_probe_tutorial.ipynb
AntoninPoche: I will see what I can do

interpreto/concepts/probes/base.py
```diff
@@ -0,0 +1,279 @@
...
+from interpreto.concepts.base import ConceptEncoderExplainer, check_fitted
+from interpreto.model_wrapping.model_with_split_points import ModelWithSplitPoints
+from interpreto.typing import ConceptsActivations, LatentActivations
+
+# ---------------------------------------------------------------------------
+# Decorator
+# ---------------------------------------------------------------------------
+
+
+def assert_fitted(fn):
```
fanny-jourdan, line 67: I'm not sure what this is for. Doesn't it already throw a runtime error by default if it doesn't work? But I guess you've already checked to make sure you need it.

interpreto/concepts/probes/base.py
```diff
@@ -0,0 +1,279 @@
...
+        """Fit the probe on activations and multi-label targets.
+
+        Args:
+            x: Latent activations.
+            y: Binary multi-label targets.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def encode(self, x: Float[torch.Tensor, "n d"]) -> Float[torch.Tensor, "n c"]:
```
fanny-jourdan, line 134: Are you sticking with “encode” as the name after all? I thought you wanted to switch to something like “activation_to_concept”?

interpreto/concepts/probes/centroid.py
```diff
@@ -0,0 +1,404 @@
...
+
+        x2: Float[torch.Tensor, "n 1"] = (xn * xn).sum(dim=1, keepdim=True)
+        c2: Float[torch.Tensor, "1 c"] = (C * C).sum(dim=1, keepdim=True).t()
+        dots: Float[torch.Tensor, "n c"] = xn @ C.t()
+        dist2: Float[torch.Tensor, "n c"] = x2 + c2 - 2.0 * dots
+        return -dist2
+
+
+class DiagonalMahalanobisCentroidProbe(BaseCentroidProbe):
+    """Centroid probe using diagonal Mahalanobis distance.
```
fanny-jourdan, line 246: Maybe if there are a paper introduced the distance you can cite it ? (remark available for all distances)

fanny-jourdan (approved):
Great job, Antonin! The code is complete and well-documented! I'll let you respond to the few comments and merge the code.

interpreto/concepts/probes/base.py
AntoninPoche: That is mainly for a clearer error message. But there might also be cases where it prevents unforeseen bugs.

interpreto/concepts/probes/base.py
AntoninPoche: It will be in a later PR, there are a lot of modifications, and I try to tackle them one by one so it's clear for reviewing. So I first did the probes, then I will rename it in a future PR.

docs/notebooks/classification_probe_tutorial.ipynb
AntoninPoche: That would indeed be interesting, but with time constraints and current examples, I will skip this.

fanny-jourdan (approved):

PR #154, opened 2026-05-27
Add vit attribution hugo

HugoDeBosschere:
## Description

Added a MVP for ViT attribution-based explanation. 

Creation of: 

1. ImageGranularity mirroring Granularity with PATCH and PIXEL granularities in interpreto/commons/granularity.py
PATCH is required for perturbation based methods (Occlusion, Sobol, Lime, KernelShap)

2. ImageClassificationInferenceWrapper in interpreto/model_wrapping/image_classification_inference_wrapper.py. 
It largely follows the implementation of InferenceWrapper and TextInferenceWrapper with the most significative changes being the following: 

- Usage of "pixel_values" instead of "input_ids"
- _compute_gradients (and thus a __call__ to ImageInferenceWrapper) returns the gradients for each pixel AND each channel compared to the text side where an aggregation is necessary for memory issues (which then means that some methods cannot be adequately implemented : VarGrad and SquareGrad)
- to follow the same pattern as the text side, instead of storing images as 2D tensors we flatten them to 1D vector where the length of the vector l = H*W where H and W are the height and width of the image.

3. GranularityResizeStrategy instead of GranularityAggregationStrategy in attributions/commons/granularity.py. It was deemed for natural for images to be able to apply different kinds of resize to smaller size explanations (because of the PATCH granularity) instead of following the discerte paradigms of the association_matrix on the text side. This allows the user to choose how he or she would want the resize to happen. If the user wants the same behavior as the text side, choosing the NEAREST interpolation has the same behavior (it maps pixel for pixel back for PATCH space to PIXEL space where each pixel in each PATCH getting the same attribution as the PATCH). A superclass linking both GranularityResizeStrategy and GranularityAggregationStrategy has been created for the future merge of Text and Image classes. 
The possible values for GranularityResizeStrategy are: 

- NEAREST = "nearest"
- BILINEAR = "bilinear"
- BICUBIC = "bicubic"
- AREA = "area"

4. ImageAttributionOutput in interpreto/attributions/base.py. This is the equivalent of AttributionOutput to store the Output of an image explanation. elements has been discarded. It was used to link the granularity elements back to the tokens elements which has been made useless by the introduction of the resize_strategy (of type GranularityResizeStrategy). ImageAttributionOutput also stores patch_size, image_mean and image_std in order to be able to denormalize the image for visualization purposes. Storing those objects allows to not store the entire raw or processed image but rather reconstruct it from model_inputs_to_explain which is needed in order to use the Insertion Deletion evaluation methods (not yet supported for images)

6.  ImageClassificationAttributionExplainer in interpreto/attributions/base.py. 
ImageAttributionExplainer is the equivalent of TextClassificationAttributionExplainer. 

Differences in the __init__ : 

- granularity_aggregation_strategy -> granularity_resize_strategy 
- granularity has been moved from optinnal argument to required so that each method has to specify which granularity it wants to use. This works because each method is linked to a unique ImageGranularity 
- We have three new arguments : preprocess, image_mean and image_std. This has been implemented to allow users to preprocess their images independently and have explanations on how the model handle their already preprocessed images. This has been done because I thought it plausible for industrials to have their own preprocessing pipeline independently from the model they are using but probably have not trained. Since this project is also aimed at them I thought it useful to add. preprocess = True means (maybe counterintuitively) that the image should be preprocessed by the model. image_mean and image_std are only useful when preprocess = False in which case we need the normalization constants to be provided by the user in order to be able to reconstruct a somewhat realistic image at visualization time. If the user does not want that, he or she can always just pass 0 and 1 as mean which is a no-op. 
Here are the new functions of this class or what has changed from the old functions: 

- _resolve_normalization_stats just verifies that image_mean and image_std have been (if preprocess = False) or have not been (if preprocess = True) passed as arguments in the __init__. It also retrieves the normalization constants from the image_processor used in the pipeline.
- process_targets and process_inputs_to_explain_and_targets are identical to their Text equivalent 
- process_model_inputs and _validate_batch_feature make different checks on the nature of the given argument to verify that the inputs are well formed.
- explain follows the same logic as its text equivalent though the dimensions do change for reasons explained in the part on ImageInferenceWrapper. The resize is not done here but rather in the viz module. 

8. Following the maintainers' advices, we now have a new Perturbator inheritance (Perturbator, TensorPerturbator, MaskPerturbator, TextTensorPerturbator, TextMaskPerturbator, ImageTensorPerturbator, ImageMaskPerturbator) though it has not yet been realized as it should because it was deemed more important to do a functionnal version with a fonctionnal test suite and only then merge the ImageMethods with the TextMethods.

9. The Aggregators have stayed the same

11. All the methods now have their Image equivalent

12. A visualization file has been written. We have chosen to draw all the explanations with the same colors so that the colors always mean the same thing (this means that for the methods that can only output positive explanations, such as VarGrad, a blue color can never be reached). The default colormap is coolwarm and by default it is shown on a grayscale image so that it is more visible. It has been chosen to show the negative contribution as negative by default.
What each function does: 

- _denormalize : denormalizes the image to get back the original image. 
- _clip_percentile : clips extreme value to preserve meaningful heatmaps 
- _prepare_heatmap : resize the explanation to PIXEL space and prepare the heatmap for one target
- _color_limits : centers the explanation on 0 by default 
-  _to_grayscale : put the background image in gray by default 
-  _draw_attribution_on_ax : draws one image and one heatmap on ax 
- plot_image_attribution : plots for several targets and several ImageAttributionOutputs the calculated explanations 

13. An extensive test suite was written to cover all the developped functions. All the tests are successful except some that may crash on gpu with not high enough memory capacity. The test files are the following: 

- test_image_granularity.py 
- test_image_attribution.py
- test_image_classification_inference_wrapper.py
- test_image_base.py 
- test_image_perturbators.py 
- test_image_methods_sanity.py (This file is a sanity check for vision methods. We put all the image but a box containing either a cat or a dog and we ask the explainer to explain the model decision on the whole image. We expect the heatmap to be contained in that box, which it is, except for KernelShap where it seems a bit more intense in the box but maybe not enough. This may be due to too small a number of perturbations though 1000 was used) 

Once it has been reviewed or before if the maintainers agree, I can begin the actual merge between all the functions to have a unified way of calling both the image and text attribution methods.

## Checklist

- [ x] I've read the [`CODE_OF_CONDUCT.md`](https://github.com/FOR-sight-ai/interpreto/blob/main/CODE_OF_CONDUCT.md) document.
- [x ] I've read the [`CONTRIBUTING.md`](https://github.com/FOR-sight-ai/interpreto/blob/main/docs/contributing.md) guide.
- [x ] I've successfully run the style checks using `make lint`.
- [x ] I've written tests for all new methods and classes that I created and successfully ran `make test`.
- [ x] I've written the docstring in Google format for all the methods and classes that I used.


interpreto/attributions/image_base.py
```diff
@@ -0,0 +1,340 @@
...
+        batch_size: int = 4,
+        perturbator: Perturbator | None = None,
+        aggregator: Aggregator | None = None,
+        device: torch.device | None = None,
+        granularity: ImageGranularity = ImageGranularity.DEFAULT,
+        granularity_aggregation_strategy: GranularityAggregationStrategy = GranularityAggregationStrategy.MEAN,
+        inference_mode: Callable[[torch.Tensor], torch.Tensor] = InferenceModes.LOGITS,
+        use_gradient: bool = False,
+        input_x_gradient: bool = True,
+        preprocess: bool = True,
```
AntoninPoche: Are you sure preprocessing by default is the way to go?

interpreto/attributions/image_base.py
```diff
@@ -0,0 +1,340 @@
...
+    model_inputs_to_explain: TensorMapping
+    targets: torch.Tensor
+    model_task: ModelTask
+    classes: torch.Tensor | None = None
+    granularity: ImageGranularity = ImageGranularity.DEFAULT
+    granularity_aggregation_strategy: GranularityAggregationStrategy = GranularityAggregationStrategy.MEAN
+    inference_mode: Callable[[torch.Tensor], torch.Tensor] = InferenceModes.LOGITS
+
+
+class ImageClassificationAttributionExplainer(ClassificationAttributionExplainer):
```
AntoninPoche: You cannot inherit directly from the `ClassificationAttributionExplainer` class. In particular, if your goal is just to reuse the target's preprocessing.

The goal is to find a way to use or adapt the `MultitaskExplainerMixin` that kind of serves as a factory. When you instantiate `Saliency`, it checks the model class and assigns the base task-explainer to inherit from both and have what you need.

The thing is that this hinders clarity. IMO, they should all be at the same level.

interpreto/attributions/image_base.py
```diff
@@ -0,0 +1,340 @@
...
+        # the internal replication is a no-op and we redo it here. Cleaner long-term: either
+        # strip the internal replication (return one decomposition) or concat samples into a
+        # single batched BatchFeature upstream and let `get_decomposition` produce the full
+        # `n_samples` list directly.
+        shared_coords: list[tuple[int, int]] = self.granularity.get_decomposition(
+            model_inputs_to_explain[0],
+            patch_size=self.patch_size,
+            return_coordinates=True,
+        )[0]  # type: ignore
+        granular_inputs_coords: list[list[tuple[int, int]]] = [shared_coords for _ in model_inputs_to_explain]
```
AntoninPoche: Not sure what you use this for.

This was used in the text to know to which words or tokens attributions correspond.

For images, you can just store the images in their original (h, w) and the attributions in the granularity (h, w). Then, for visualizations, we resize (which does nothing for pixels and something for patches). (Note that the way of resizing can be a visualization parameter.)

In summary, if you already store the image, I would set `None` for the elements or omit it.

interpreto/attributions/image_base.py
```diff
@@ -0,0 +1,340 @@
...
+
+        results: list[ImageAttributionOutput] = []
+        for contribution, model_input, elements, target in zip(
+            granular_contributions,
+            model_inputs_to_explain,
+            granular_inputs_coords,
+            sanitized_targets,
+            strict=True,
+        ):
+            model_task, clean_contribution = self.post_processing(contribution)
```
AntoninPoche: You do not implement `self.post_processing`. However, the parent class just does `return ModelTask.CLASSIFICATION, contribution`.

I do not think you should use the same `ModelTask`, at least for now.

So you can remove this line and just create an `ImageClassification` task to pass to your `ImageAttributionOutput`. Or just omit it.

If you omit everything you do not need, it might be easier to identify what's common and not for later merging.

interpreto/commons/image_granularity.py
```diff
@@ -0,0 +1,271 @@
...
+                # rows of patches, then cols of patches
+                for row_start in range(0, h, patch_size):
+                    for col_start in range(0, w, patch_size):
+                        patch = []
+                        for i in range(patch_size):
+                            for j in range(patch_size):
+                                # pixel positions in this patch, starting from row_start and col_start and moving by patch_size pixels
+                                patch.append((i+row_start) * w + j + col_start)
+                                
+                        per_sample.append(patch)
```
AntoninPoche: There should be a simpler solution than 4 nested for loops.

In any case, my first intuition is that you do not the `get_indices`. At least for the two granularities you have here.

The granularity is used at two steps:
- For perturbations, either your perturb the pixel_values or the patch so no need for granularity.
- For Aggregation if your attributions are not already at the correct granularity. In this case, for now, you just have the case were you request patch for gradients where you need to downscale the attributions from pixel-wise to patch-wise.

So IMO, you do not need most of the granularity elements for images. It is useful for text because each sample has a different size, with words and sentences distributed across the whole sample. It is much more consistent with images.

Well, the above comments do not make sense if you have special patch/tokens which do not correspond to pixels but to constants (to my understanding, these are used in ViTs and such).

interpreto/commons/image_granularity.py
```diff
@@ -0,0 +1,271 @@
...
+                    assert unit_contribution.dim() == 2 and unit_contribution.shape[1] >= 2, (
+                        f"PATCH units must contain >=2 pixels; got shape {tuple(unit_contribution.shape)}. "
+                        "Use ImageGranularity.PIXEL for pixel-level granularity."
+                    )
+                    aggregated_contribution[:, [j]] = granularity_aggregation_strategy.aggregate(
+                        unit_contribution, dim=1
+                    )
+                return aggregated_contribution
+            case _:
+                raise NotImplementedError(f"Granularity {self} not implemented for aggregation.")
```
AntoninPoche: I imagine that this should be not be based on indices but rather on interpolations:
https://docs.pytorch.org/docs/2.12/generated/torch.nn.functional.interpolate.html

interpreto/commons/image_granularity.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def get_decomposition(
+        self,
+        inputs: TensorMapping,
+        patch_size: int = 16,
+        return_coordinates: bool = False,
+        indices_list: list[list[list[int]]] | None = None,
+        **kwargs,
+    ) -> list[list[list[int]]] | list[list[tuple[int, int]]]:
```
AntoninPoche: If my comment on `image_base.py` is true, this can be removed.

interpreto/model_wrapping/image_classification_inference_wrapper.py
```diff
@@ -0,0 +1,173 @@
...
+
+            # input×gradient: same scheme as text, just multiplying against pixel_values_chunk
+            # instead of inputs_embeds_chunk
+            if self.input_x_gradient:
+                target_wise_grads = target_wise_grads * pixel_values_chunk
+
+            # text does `.abs().mean(dim=-1)` to collapse the d=768/4096 width dim for memory;
+            # here we collapse channels for the 1D l contract (not memory) and flatten H,W to l=H*W
+            target_wise_mean_grads: Float[torch.Tensor, f"{c} l"] = (
+                target_wise_grads.abs().mean(dim=1).flatten(start_dim=1)
```
AntoninPoche: Why flatten? I know it breaks the shape from text, but it makes much more sense to have an (H,W) shape IMO.

This depends on how the rest is done.

interpreto/attributions/image_base.py
AntoninPoche: But for now, it can be kept like this if you want easy tests.

interpreto/attributions/methods/image_saliency.py
```diff
@@ -0,0 +1,128 @@
...
+        >>> explanations = method.explain(image)
+    """
+
+    def __init__(
+        self,
+        model: PreTrainedModel,
+        image_processor: BaseImageProcessor,
+        batch_size: int = 4,
+        granularity: ImageGranularity = ImageGranularity.DEFAULT,
+        granularity_aggregation_strategy: GranularityAggregationStrategy = GranularityAggregationStrategy.MEAN,
```
AntoninPoche: I imagine that for images, this should correspond to the interpolation mode.

pyproject.toml
```diff
@@ -89,6 +89,7 @@ dependencies = [
   "nvidia-cusparse-cu11>=11.7.4.91; sys_platform=='Linux'",
   "nvidia-nccl-cu11>=2.14.3; sys_platform=='Linux'",
   "nvidia-nvtx-cu11>=11.7.91; sys_platform=='Linux'",
+  "torchvision>=0.27.0",
```
AntoninPoche, line 92: We should discuss with @fanny-jourdan  if we want `torchvision` to be a main dependency or in something like `pip install interpreto[vision]`.

interpreto/attributions/image_base.py
```diff
@@ -78,6 +78,13 @@ class ImageAttributionOutput:
             The output of `image_processor(image, return_tensors="pt")` (a `BatchFeature`,
             satisfies `TensorMapping`). Holds `pixel_values` of shape `(1, 3, H, W)`.
 
+        raw_image (PIL.Image | np.ndarray | torch.Tensor | None):
```
AntoninPoche: I think it should be harmonized before being saved. It will be much easier for visualizations.

So it should have a single type.

interpreto/attributions/image_base.py
```diff
@@ -267,6 +275,21 @@ def explain(
...
+        if isinstance(model_inputs, list):
+            raw_images = [
+                m if self.preprocess and isinstance(m, (PILImage, np.ndarray, torch.Tensor)) else None
+                for m in model_inputs
+            ]
+        elif self.preprocess and isinstance(model_inputs, (PILImage, np.ndarray, torch.Tensor)):
+            raw_images = [model_inputs]
+        else:
+            raw_images = [None] * len(model_inputs_to_explain)
+
```
AntoninPoche: I think this should go in `process_model_inputs`. So you know that you return None if inputs are already processed.

AntoninPoche (changes requested):
The commit messages are clear, so I can review them commit by commit. Its nice! I guess thanks, Claude ^^

Overall, it is clear, and it is nice that you already managed to have something running. My main comment is that ImageGranularity is far too complex for what you need for images. But you will really know after trying perturbations at both granularities.

In any case, you can discuss all comments; they are my opinions, not based on having tried the thing myself.

interpreto/attributions/image_base.py
AntoninPoche: We discussed and this outdated

interpreto/model_wrapping/image_classification_inference_wrapper.py
AntoninPoche: You told it was for aggregation right?

interpreto/attributions/perturbations/image_base.py
AntoninPoche, line 1: I think you only need the MaskPerturbator one. But we will discuss with Thomas how to organize the whole thing.

interpreto/commons/image_granularity.py
AntoninPoche: Partially outdated

interpreto/commons/image_granularity.py
AntoninPoche: Also outdated, but still, the interpolation is definitely a useful operation!

In particular, when rescaling with bilinear interpolation, the results are smooth. This is used in Rise and GradCAM and makes these methods much more pleasant to explainee.

AntoninPoche:
Hi @HugoDeBosschere, with @thomas-mullor we discussed how to tackle the perturbators problem.

@fanny-jourdan do not hesitate to give your opinion.

To summarize, the problem was that perturbators depend both on the method and the modality.
But we do not want to implement each combination of method and modality.

So we found a way to dynamically create the necessary perturbator by inheriting from both the method and modality specific classes.

Here is a summary of the modifications:

```python
# attributions/perturbations/base.py

class Perturbator(ABC):
    @abstractmethod
    def perturb(self, inputs):
        pass

class TensorPerturbator(Perturbator):  # new class (just for typing and clarity)
    @abstractmethod
    def perturb_tensor(self):  # renaming of `perturb_embeds`
        pass

class TextTensorPerturbator(TensorPerturbator):  # renaming of `EmbeddingsPerturbator`
    def perturb(self, inputs):  # already exists
        # calls `perturb_tensor`

class ImageTensorPerturbator(TensorPerturbator):  # renaming of your `ImageEmbeddingsPerturbator`
    def perturb(self, inputs):  # already exists but can surely be simplified
        # calls `perturb_tensor`

class MaskPerturbator(Perturbator):  # new class (just for typing and clarity)
    @abstractmethod
    def get_mask(self):
        pass

class TextMaskPerturbator(MaskPerturbator):  # renaming of `IdsPerturbator`
    def perturb(self, inputs):  # already exists
        # calls `get_mask`

class ImageMaskPerturbator(MaskPerturbator):  # renaming of your `ImageIdsPerturbator`
    def perturb(self, inputs):  # already exists
        # calls `get_mask`
```

```python
# attributions/perturbations/occlusion_perturbation.py

from .base import MaskPerturbator

class OcclusionPerturbator(MaskPerturbator):  # change inheritance, might make the type checker unhappy
    def get_mask(self, mask_dim):
        # should support both `TextMaskPerturbator` and `ImageMaskPerturbator` requirements
        # it might be nothing at first
```

```python
# attributions/perturbations/gaussian_noise_perturbation.py

from .base import TensorPerturbator

class GaussianNoisePerturbator(TensorPerturbator):  # change inheritance, might make the type checker unhappy
    def perturb_tensor(self, inputs_embeds):
        # should support both `TextTensorPerturbator` and `ImageTensorPerturbator` requirements
        # therefore, support (1, l, d) and (1, c, h, w) shapes
        # (1, ...) -> (p, ...)
```

```python
# attributions/base.py

from .perturbations.base import (
    Perturbator,
    TensorPerturbator,
    TextTensorPerturbator,
    ImageTensorPerturbator,
    MaskPerturbator,
    TextMaskPerturbator,
    ImageMaskPerturbator
)

class AttributionExplainer(ABC):
    associated_inference_wrapper: InferenceWrapper
    base_tensor_perturbator_class: type[TensorPerturbator]
    base_mask_perturbator_class: type[MaskPerturbator]

    # does not impact the methods

class TextClassificationExplainer(AttributionExplainer):
    associated_inference_wrapper: TextClassificationInferenceWrapper
    base_tensor_perturbator_class = TextTensorPerturbator
    base_mask_perturbator_class = TextMaskPerturbator

class TextGenerationExplainer(AttributionExplainer):
    associated_inference_wrapper: TextGenerationInferenceWrapper
    base_tensor_perturbator_class = TextTensorPerturbator
    base_mask_perturbator_class = TextMaskPerturbator

class ImageClassificationExplainer(AttributionExplainer):
    associated_inference_wrapper: ImageClassificationInferenceWrapper
    base_tensor_perturbator_class = ImageTensorPerturbator
    base_mask_perturbator_class = ImageMaskPerturbator

class MultitaskExplainerMixin:
    # no modifications specific to the new perturbator classes
    # still, it should include the ImageClassificationExplainer at some point
```

```python
# attributions/methods/occlusion.py

class Occlusion(MultitaskExplainerMixin, AttributionExplainer):
    def __init__(...):
        # create the perturbator dynamically by inheriting from both the method and modality specific classes
        perturbator_class = type(
            "ModalitySpecific" + self.__class__.__name__,  # name
            (OcclusionPerturbator, self.base_mask_perturbator_class,),  # parent classes
            {}
        )
        perturbator = perturbator_class(
            tokenizer=tokenizer,
            granularity=granularity,
            replace_token_id=replace_token_id,
        )
        ...
```

```python
# attributions/methods/smooth_grad.py

class SmoothGrad(MultitaskExplainerMixin, AttributionExplainer):
    def __init__(...):
        # create the perturbator dynamically by inheriting from both the method and modality specific classes
        perturbator_class = type(
            "ModalitySpecific" + self.__class__.__name__,  # name
            (GaussianNoisePerturbator, self.base_tensor_perturbator_class,),  # parent classes
            {}
        )
        perturbator = perturbator_class(
            inputs_embedder=model.get_input_embeddings(),
            n_perturbations=n_perturbations,
            std=noise_std
        )
        ...
```

interpreto/attributions/image_base.py
HugoDeBosschere: I would have said yes because in my head the typical use case would be using this on a PIL image where this requires preprocessing. Of course maybe I'm wrong and it will be more generally used in already preprocessed tensors in which case it may be better to put preprocess = False. 
I think this has to be resolved by how it will most often be used, which I have little visibility on.

interpreto/attributions/image_base.py
HugoDeBosschere: Okay I removed it.

interpreto/model_wrapping/image_classification_inference_wrapper.py
HugoDeBosschere: This allows us to reuse aggregator entirely without changing anything. It also allowed me to copy with minimal changes a lot of the functions that were already written for text (such as the perturbators for example, or the granularity) so that we may more easily refactor after.

interpreto/attributions/methods/image_saliency.py
HugoDeBosschere: I'm not sure if I correctly understand the comment but granularity_aggregation_strategy is the strategy used when use_gradient = True to go from pixel scores to granularity scores. 

If the interpolation mode you reference here is the one in the aggregator then, from what I understand, this aggregates on the perturbations, not the granularity. 

So I'm not sure about what to respond.

interpreto/attributions/image_base.py
HugoDeBosschere: I had kept it this way because I didn't want to plot a modified image for the user (I thought it was best if he saw the same image he had first loaded). I already did the visualization script so : do I still need to modify this field ?
If the concern was that the visualization was going to be hard then maybe I can keep it. 
If it was simplicity then I can just use model_inputs_to_explain as the base image onto which to display the heatmap.

interpreto/attributions/image_base.py
AntoninPoche: What happens if we process already processed inputs? Does it do something weird?

If so, then it might be tricky, otherwise, default preprocess is the way to go.

interpreto/model_wrapping/image_classification_inference_wrapper.py
AntoninPoche: Then it is a good choice!

interpreto/attributions/methods/image_saliency.py
AntoninPoche: I am talking about the granularity of the aggregation here.

After gradients on the pixel_values, if you want to go to the patch granularity, you need to resize the explanations.

The image resizing operation takes an interpolation parameter. This parameter is a kind of choice we make with the GranularityAggregationStrategy. Well, you could also call this pooling, but resize allows you to think at the image level.

interpreto/attributions/image_base.py
AntoninPoche: What you are saying is that once processed, if we plot the image, it looks weird, so we need to keep a raw image, right?

If that is the case, I agree with you.

interpreto/attributions/image_base.py
HugoDeBosschere: In most cases it will do something weird indeed. But if the user does not preprocess the inputs it also does something weird (the ViT will not function as expected). So I don't know which one is best.

interpreto/attributions/methods/image_saliency.py
HugoDeBosschere: Okay I undestand better.
Interpolation and pooling both work but I have a preference for pooling since this looks a lot like your typical max pooling operation (when the strategy is MAX) and those terms are generally used in CNNs (here it would be a patch_size * patch_size pooling operation). If you think interpolation is better I can put interpolation.

interpreto/attributions/image_base.py
AntoninPoche: The current behavior looks good to me. It is just that I was not familiar with how things work in ViTs, and I wanted the default choice to preprocess or not to be made consciously.

interpreto/attributions/methods/image_saliency.py
AntoninPoche: I think resizing is better adapted, indeed, because it considers the image as a whole. While pooling makes small windows and extracts a value from them.

Here, taking the maximum, mean, or any other value does not really make sense. Well, it kind of does the job, but resizing is a much better way to change an image's size. (Which also applies to attributions with shape (h, w)).

Some methods, like the CAM family or Rise, provide smooth, human-friendly attributions as a result.

HugoDeBosschere: Ok I'll look into replacing GranularityAggregationStrategy with an InterpolationStrategy and to define several stategies of interpolation

interpreto/attributions/image_base.py
HugoDeBosschere: I haven't actually tried to plot it with the processed image. Since the comment had not been submitted before I actually wrote the visualization functions, I used the raw_image and it works. Do you want me to try and see if I can also make it work with the processed image so that we may drop this field that may not be useful ?

interpreto/attributions/perturbations/base.py
```diff
@@ -28,51 +28,42 @@
...
-        """
-        # add perturbation dimension
-        if model_inputs["input_ids"].ndim <= 1:
-            model_inputs["input_ids"] = model_inputs["input_ids"].unsqueeze(0)
-        # model_inputs["input_ids"] = model_inputs["input_ids"].unsqueeze(0)
-        # if "inputs_embeds" in model_inputs:
-        #    model_inputs["inputs_embeds"] = model_inputs["inputs_embeds"].unsqueeze(0)
-        # TODO : eventually add perturbation dimension to other keys in the mapping ?
-
-        return model_inputs, torch.zeros_like(model_inputs["input_ids"], dtype=torch.float)
```
fanny-jourdan, line 69: Le comportement de base du perturb il disparait? Je ne crois pas l'avoir vu tel quel dans une classe de base de la partie texte ?

interpreto/attributions/perturbations/base.py
```diff
@@ -265,3 +262,194 @@ def perturb(self, model_inputs: TensorMapping) -> tuple[TensorMapping, torch.Ten
...
+        Returns:
+            perturbed_embeds: Shape (p, 3, H, W).
+            mask: (p, g) or None.
+        """
+        return pixel_values, None
+
+
+class ImageMaskPerturbator(MaskPerturbator):
+    """
+    Image-side analog of `IdsPerturbator`.
```
fanny-jourdan, line 314: il n'y a plus de IdsPerturbator, tu l'as renommé TextMaskPerturbator je crois

fanny-jourdan: (Il apparait à beaucoup d'endroit, je te laisse changer partout)

interpreto/attributions/perturbations/shap_perturbation.py
```diff
@@ -120,3 +120,78 @@ def get_mask(self, mask_dim: int) -> Float[Tensor, "p {mask_dim}"]:
...
+        )
+        self.device = device  # type: ignore
+
+    @jaxtyped(typechecker=beartype)
+    def get_mask(self, mask_dim: int) -> Float[Tensor, "p {mask_dim}"]:
+        """
+        Sample binary masks weighted by the Shapley kernel.
+
+        See `ShapTokenPerturbator.get_mask` for the full derivation; the sampling
+        is modality-agnostic and operates only on the number of units `mask_dim`.
```
fanny-jourdan, line 162: je n'ai pas trop comprit le code de ton get_mask ici, est ce que tu pourrais m'expliquer un peu plus en details ce que tu fais s'il te plait?

interpreto/attributions/perturbations/sobol_perturbation.py
```diff
@@ -110,3 +110,65 @@ def get_mask(self, mask_dim: int, **kwargs) -> Float[torch.Tensor, "p {mask_dim}
...
+            granularity=granularity,
+            n_perturbations=-1,
+            replace_value=replace_value,
+            patch_size=patch_size,
+        )
+        self.n_token_perturbations = n_token_perturbations
+        self.sampler_class = sampler.value
+
+    @jaxtyped(typechecker=beartype)
+    def get_mask(self, mask_dim: int, **kwargs) -> Float[torch.Tensor, "p {mask_dim}"]:
```
fanny-jourdan, line 149: Est ce que tu utilises le code de xplique ici ?

tests/model_wrapping/test_classification_inference_wrapper.py
fanny-jourdan, line 1: Vu que tu as renommé ClassificationInferenceWrapper ça serait bien de renommé le test, test_text_classification_inference_wrapper.py aussi. Cette remarque vaut pour tout les autres tests.

interpreto/model_wrapping/classification_inference_wrapper.py
fanny-jourdan, line 1: Il faudrait renommer le fichier text_classification_inference_wrapper.py du coup (remarque valable pour toutes les autres classes du genre)

pyproject.toml
fanny-jourdan: Oui je pense que ça ne devrait pas etre une dépendance par défaut de la lib

.pre-commit-config.yaml
```diff
@@ -24,7 +24,7 @@ repos:
         exclude: LICENSE
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.10
+    rev: v0.15.22
```
fanny-jourdan, line 27: quand on fera le changement de version pour ajouter la vision on passera à la 0.16 je pense

interpreto/commons/granularity.py
```diff
@@ -903,3 +979,127 @@ def granularity_score_aggregation(  # noqa: PLR0912  # ignore too many branches
...
+
+    Positions are flattened row-major: index `i` <-> `(row=i//W, col=i%W)`.
+    For `vit-base-patch16-224`: PIXEL has `H*W = 50176` singleton units,
+    PATCH has `196` units of `patch_size*patch_size = 256` pixel positions each.
+    """
+
+    PIXEL = "pixel"
+    PATCH = "patch"
+
+    #    def granularity_resize(
```
fanny-jourdan, line 1001: Beaucoup de chose commenté # à partir de là, peut etre que tu devrais faire une pass pour clean le code dans le fichier granularity.py si tu ne t'en sers pas

interpreto/attributions/base.py
```diff
@@ -29,35 +29,52 @@
...
+from interpreto.attributions.perturbations.base import (
+    ImageMaskPerturbator,
+    ImageTensorPerturbator,
+    MaskPerturbator,
+    Perturbator,
+    TensorPerturbator,
+    TextMaskPerturbator,
+    TextTensorPerturbator,
+)
+from interpreto.commons import TextGranularity
```
fanny-jourdan, line 56: il faudrait que tu load le ImageGranularity de la meme façon que le TextGranularity (je pense que ça doit etre un truc à ajouter dans un fichier de commons)

interpreto/attributions/base.py
```diff
@@ -29,35 +29,52 @@
...
 from interpreto.model_wrapping.inference_wrapper import InferenceModes, InferenceWrapper
 from interpreto.typing import ClassificationTarget, GeneratedTarget, ModelInputs, SingleAttribution, TensorMapping
 
 
-def setup_token_ids(model: PreTrainedModel, tokenizer: PreTrainedTokenizer, require_mask_token: bool = True) -> int:
+def setup_token_ids(
+    model: PreTrainedModel, tokenizer: PreTrainedTokenizer | BaseImageProcessor, require_mask_token: bool = True
+) -> int | None:
     """
     Setup the tokenizer and the model with the appropriate token IDs, for padding and masking.
```
fanny-jourdan, line 70: change the description if you add image processor on this

interpreto/attributions/base.py
```diff
@@ -404,26 +424,20 @@ def process_inputs_to_explain_and_targets(
...
         Returns:
             model_task (ModelTask): The model task.
             contribution (Float[torch.Tensor, "t l"]): The post-processed contribution values.
+
         """
-        raise NotImplementedError(
-            "Specific task subclasses must implement the 'post_processing' method "
-            "to correctly post-process the contribution values."
-        )
+        return contribution
```
fanny-jourdan, line 440: Je ne comprends pas tes changements ici. On avait une fonction abstraite qui était instancier proprement dans les sous classes, je ne vois pas pourquoi tu as fait ça a la place?

interpreto/attributions/base.py
```diff
@@ -680,30 +698,12 @@ def process_inputs_to_explain_and_targets(
...
 
-
-class GenerationAttributionExplainer(AttributionExplainer):
-    """
-    Attribution explainer for generation models
-    """
-
-    _associated_inference_wrapper = GenerationInferenceWrapper
-    inference_wrapper: GenerationInferenceWrapper
+class TextGenerationExplainer(AttributionExplainer):
```
fanny-jourdan, line 702: Better to keep TextGenerationAttributionExplainer puisqu'on a aussi d'autre type d'explainer dans la lib

interpreto/attributions/base.py
```diff
@@ -680,30 +698,12 @@ def process_inputs_to_explain_and_targets(
 
         return model_inputs, sanitized_targets
 
-    def post_processing(self, contribution: Float[torch.Tensor, "t l"]):
```
fanny-jourdan, line 683: Cette fonction devrait rester ici avec la forme de base. C'est plus propre que d'avoir une version implémenter dans la classe abstract comme tu as fait là

interpreto/attributions/base.py
```diff
@@ -843,10 +843,545 @@ def post_processing(self, contribution: Float[torch.Tensor, "t l"]):
         t, l = contribution.shape
         mask = torch.triu(torch.ones((t, l), dtype=torch.bool), diagonal=l - t)
         contribution[mask] = float("nan")
-        return ModelTask.GENERATION, contribution
+        return contribution
```
fanny-jourdan, line 846: Je comprends pas pourquoi tu as retirer l'info sur la tache de partout

interpreto/attributions/base.py
```diff
@@ -843,10 +843,545 @@ def post_processing(self, contribution: Float[torch.Tensor, "t l"]):
...
+            Canonical per-unit attribution scores of shape `(t, g)`, where `g = H*W` (PIXEL)
+            or `g = num_patches` (PATCH). This is the lossless numeric output — one scalar per
+            granularity unit. Stored FLAT; patch `(row, col)` coordinates are derivable on demand
+            from `model_inputs_to_explain["pixel_values"].shape[-2:]` and `patch_size`.
+
+        patch_size (int):
+            Patch side length, used to lay out PATCH-granularity scores on the pixel grid
+            (`gh, gw = H // patch_size, W // patch_size`). Stored so the visualization can
+            rebuild the display-ready `(t, H, W)` map from `attributions` via `resize_to_image`.
+            Ignored for PIXEL granularity, where `g == H*W` and the expansion is a pure reshape.
```
fanny-jourdan, line 872: On n'a pas une valeur par défaut ici ?

interpreto/attributions/base.py
```diff
@@ -843,10 +843,545 @@ def post_processing(self, contribution: Float[torch.Tensor, "t l"]):
...
+
+        # Granularity is a strict function of the method family, so the two arguments cannot
+        # be chosen independently. Gradient methods differentiate with respect to every pixel and only
+        # afterwards resize the field, so PIXEL is their native unit; perturbation methods
+        # encode granularity in the masks themselves so PATCH is the smallest meaningful unit
+        if use_gradient and granularity is ImageGranularity.PATCH:
+            raise ValueError(
+                "granularity=PATCH is invalid for a gradient-based method: gradients are "
+                "computed per pixel, so PIXEL is the granularity these methods explain at."
+            )
```
fanny-jourdan, line 996: Fais le meme message d'erreur que tu avait fait dans l'autre sens avec un "Use PIXEL." à la fin, comme ça c'est plus uniforme

tests/attributions/methods/test_attribution_methods_vision.py
```diff
@@ -0,0 +1,543 @@
...
+
+plt.switch_backend("Agg")  # headless: render into a buffer, never open a window.
+
+DEVICE = "cuda" if torch.cuda.is_available() else "cpu"
+
+CLASSIFICATION_MODELS = [
+    "hf-internal-testing/tiny-random-vit",
+    "hf-internal-testing/tiny-random-BeitForImageClassification",
+    "hf-internal-testing/tiny-random-ViTForImageClassification",
+]
```
fanny-jourdan, line 85: Pas sure de moi mais je pense que c'est mieux si tu nommes ça "IMAGE_CLASSIFICATION_MODELS" car si y'a plusieurs trucs de tests qui sont nommé pareil dans différents fichiers ça peut faire des bug bizarres de souvenir

tests/attributions/methods/test_attribution_methods_vision.py
fanny-jourdan, line 1: Quand tu mets un assert il faut ensuite qu'il retourne quelque chose si il y a une erreur pour savoir quoi changer. Par exemple `assert A==B , "A must have the same value as B"`

Il faudrait que tu ajoutes cela à tout tes assert sur tout les tests (oui désolé ça va etre un peu long mais pour la longévité du code c'est assez important car si on a une erreur quelque part quand on modifiera du code il faut qu'on puisse avoir direct l'info qui nous pose probleme)

tests/attributions/test_image_base.py
```diff
@@ -0,0 +1,397 @@
...
+    # (-1, 1, 1) so a scalar broadcasts over C and a per-channel stat aligns with it.
+    assert mean.shape == (len(expected_mean), 1, 1)
+    assert std.shape == (len(expected_std), 1, 1)
+    assert mean.dtype is torch.float32
+    assert std.dtype is torch.float32
+    assert torch.allclose(mean.flatten(), torch.tensor(expected_mean))
+    assert torch.allclose(std.flatten(), torch.tensor(expected_std))
+
+
+# NOTE FOR REVIEWERS: test_inference_mode and test_process_targets below are copied
```
fanny-jourdan, line 151: Dans cette version là ça va bien car on a volontairement pas essayé de merge la vision et le texte, mais dans la version finale il faudra penser à aussi tenter de merge les tests qu'on peut merge. :)

tests/visualizations/test_image_attribution.py
```diff
@@ -0,0 +1,331 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""
+Test file to test the visualization functions
```
fanny-jourdan, line 26: C'est pas très clair comme nom de fichier du coup si c'est fait pour tester les vizu

tests/test_image_granularity.py
```diff
@@ -0,0 +1,224 @@
...
+    strategy = GranularityResizeStrategy.AREA
+    print(strategy.resize(matrix, output_size=(2, 2)))
+    bool_tensor = strategy.resize(matrix, output_size=(2, 2)) - torch.tensor([[0.75, 0.0], [0.0, 1.75]])
+    print(bool_tensor)
+    bool_tensor = bool_tensor.abs() <= 10 ** (-5)
+    print(bool_tensor)
+    assert bool_tensor[0, 0, 0] and bool_tensor[0, 0, 1] and bool_tensor[0, 1, 0] and bool_tensor[0, 1, 1]
+
+
+# NOTE: this used to be a test for the now useless granularity_resize function. We keep it for now in case it proves useful later
```
fanny-jourdan, line 153: Finalement tu comptes t'en servir ou pas du coup ?

fanny-jourdan (commented):
Great job, Hugo, well done!!
Take a look at my comments and go over the few things I asked you to fix, even though overall it's very good. 👍

tests/attributions/methods/test_attribution_methods_vision.py
fanny-jourdan: (Ce commentaire vaut pour tout les autres fichiers test dans lequel tu as introduits CLASSIFICATION_MODELS)

.pre-commit-config.yaml
fanny-jourdan: En fait je dis ça mais là c'est pas une PR sur le main donc pas besoin de ça je pense, on le fera à la fin lors du merge final

interpreto/attributions/perturbations/base.py
HugoDeBosschere: C'est fait

interpreto/attributions/perturbations/shap_perturbation.py
HugoDeBosschere: Okay donc le truc c'est que je pense que le sampling fait dans ShapTokenPerturbator n'est pas correct au sens où il ne suit pas : 
- Sa propre documentation 
- La manière d'implémenter ce sampling dans Captum
- La formule du papier qui a introduit KernelSHAP 

Voici le docstring de ShapTokenPerturbator. 

`The perturbed instances are sampled that way:
         - We choose a number of selected features k, considering the distribution
                p(k) = (nb_features - 1) / (k * (nb_features - k))
            where nb_features is the total number of features in the interpretable space
         - Then we randomly select a binary vector with k ones, all the possible sample
           are equally likely. It is done by generating a random vector with values drawn
           from a normal distribution and keeping the top k elements which then will be 1
           and other values are 0.
         Since there are nb_features choose k vectors with k ones, this weighted sampling
         is equivalent to applying the Shapley kernel for the sample weight, defined as:
            k(nb_features, k) = (nb_features - 1)/(k*(nb_features - k)*(nb_features choose k))`

On veut donc sampler des vecteurs avec k "1" (les dimensions qui seront ensuite masquées) en suivant la formule du kernel de KernelShap c'est-à-dire en attribuant une probabilité :

` k(nb_features, k) = (nb_features - 1)/(k*(nb_features - k)*(nb_features choose k))` 

à chaque vecteur ayant n dimensions et k "1" (donc n-k "0"). 

Une manière équivalente de faire ça (c'est le "trick" dont parle le docstring et celui utilisé dans Captum) est de d'abord sampler la probabilité d'avoir k "1" en utilisant la probabilité :

`p(k) = (nb_features - 1) / (k * (nb_features - k))` 

puis de sampler de manière equiprobable n'importe quel vecteur avec k "1". (C'est une loi uniforme discrète avec k parmi n (n choose k) éléments) 

C'est bien ce qui est fait dans ShapImagePerturbator: 

`probability_to_select_k_elements: Float[Tensor, f"{l - 1}"] = (l - 1) / (possible_k * (l - possible_k))`

(l = n, possible_k = k)
Ce n'est pas ce qui est fait dans ShapTokenPerturbator: 

`probability_to_select_k_elements: Float[Tensor, f"{l - 1}"] = (possible_k * (l - possible_k)) / (l * (l - 1))`

Cette justification est donnée en commentaire pour ne pas utiliser la bonne formule: 

`# initially: (l - 1) / (possible_k * (l - possible_k)), but it gave a weird distribution` 

Le lien du papier qui introduit KernelSHAP pour que tu puisses vérifier ce que je dis 

[https://arxiv.org/pdf/1705.07874](url) en haut de la page 6 pour la définition du kernel

Je pense que mon get_mask est donc correct et celui de ShapTokenPerturbator ne l'est pas et mériterait d'être modifié.

interpreto/attributions/perturbations/sobol_perturbation.py
HugoDeBosschere: Non c'est juste un copier-collé de SobolTokenPerturbator. 
Je me disais qu'en applatissant l'image de taille (W,H) -> (l,) je peux réutiliser tel quel le code pour la perturbation. 

En fait c'est un problème parce que le papier qui introduit Sobol utilise des masques continu donc l'implémentation avec des masques binaires, qui marche pour du NLP, n'est pas la bonne pour des images. 
Il faut que je réflechisse à changer ça mais ça m'embête parce que par la suite il faudra fusionner SobolImagePerturbator et SobolTextPerturbator en SobolPerturbator donc le fait que le code soit le même m'arrangeait grandement.

Je vais donc y réflechir et voir comment le merge est possible par la suite.

interpreto/model_wrapping/text_classification_inference_wrapper.py
HugoDeBosschere: C'est fait. Je n'ai pas rennomé les methodes (type occlusion en text_occlusion alors que image_occlusion existe) parce que les méthodes seront de toute façon merge (ie occlusion (implicitement text_occlusion) et image_occlusion -> occlusion_merge (qui sera renommé occlusion tout court))

.pre-commit-config.yaml
HugoDeBosschere: Donc je repasse à la version précédente ?

interpreto/commons/granularity.py
HugoDeBosschere: C'est du code qui n'est effectivement plus utile / plus utilisé dans la librairie. 
ça servait avant quand on voulait pouvoir faire de la granularité PATCH pour les méthodes basées gradients (on a dit que l'on faisait méthode basée gradient = PIXEL, méthode basée masque = PATCH). 

Je le gardais si on change d'avis dans le futur. Si tu penses que ça ne sert à rien je peux l'enlever.

interpreto/attributions/base.py
HugoDeBosschere: C'est fait

interpreto/attributions/base.py
HugoDeBosschere: This changes in the following PR so I removed the possibility of tokenizer being of type BaseImageProcessor.

interpreto/attributions/base.py
HugoDeBosschere: C'est fait

interpreto/attributions/base.py
HugoDeBosschere: La raison pour laquelle j'ai fait ça c'est que cette fonction est la même pour les deux ClassifciationAttributionExplainer. 

J'avais l'impression de simplifier le code en faisant ça. 

Si tu souhaites quand même que je laisse la fonction sous forme abstraite dans la classe abstraite je peux.

interpreto/attributions/base.py
HugoDeBosschere: Parce que j'ai ajouté un attribut de classe _model_task associé à chaque AttributionExplainer et que la seule utilié de ce return était de récupérer la Task pour pouvoir remplir/populer (? populate) AttributionOutput. 

Puisque ce n'est plus utile, je l'ai retiré.

interpreto/attributions/base.py
HugoDeBosschere: La valeur par défaut est récupérée de model.config à la ligne 1024 du même fichier : 

`self.patch_size = int(getattr(model.config, "patch_size", 16))`

J'ai mis 16 comme valeur par défaut.

interpreto/attributions/base.py
HugoDeBosschere: C'est fait

tests/attributions/methods/test_attribution_methods_vision.py
HugoDeBosschere: J'ai changé dans tous les fichiers liés à l'image mais du coup le bug ne pourrait-t-il pas avoir lieu entre les fichiers images eux-mêmes ?

tests/visualizations/test_image_attribution.py
HugoDeBosschere: C'est vrai mais j'ai juste copié le nom du fichier qui testait les vizu côté texte (test_attributions.py que j'ai maintenant renommé test_text_attributions.py en suivant une autre suggestion) 

Je peux renommer ça différemment si tu veux.

tests/test_image_granularity.py
HugoDeBosschere: Pour le moment non mais peut-être dans le futur si on veut réintroduire la possibilité d'avoir la granularité PATCH pour les méthodes basées gradients. 

Qu'est-ce que tu en penses ? Je peux l'enlever si tu veux.

interpreto/attributions/image_base.py
HugoDeBosschere: So in the end I actually chose to go against this and we now do not store the image in ImageAttributionOutput but instead the pre-processing values (mean and std)

tests/attributions/methods/test_attribution_methods_vision.py
HugoDeBosschere: C'est fait ! 

Par contre il n'y a pas d'équivalent pour les pytest.raises / pytest.warns (ie je ne crois pas pouvoir ajouter un message précis de pourquoi est-ce que ça échoue sauf peut-être en introduisant des structures très lourdes type try/except autours de chaque pytest.raises). 

Est-ce que laisser les pytest.raises comme ça est suffisant ?

tests/model_wrapping/test_text_classification_inference_wrapper.py
HugoDeBosschere: C'est fait

interpreto/attributions/base.py
HugoDeBosschere: J'explique la raison dans la réponse au prochain commentaire sur le même sujet.

interpreto/attributions/perturbations/base.py
HugoDeBosschere: En fait c'est juste que Perturbator devient une ABC donc perturb devient une méthode abstraite. 
Puisque perturb a 4 formes différentes en fonction de la classe dans laquelle il est appelé (TextTensorPerturbator, TextMaskPerturbator, ImageTensorPerturbator, ImageMaskPerturbator) c'est assez logique de faire ça. 

D'ailleurs je suis allé regarder sur le main et le comportement de base du perturb n'était en fait jamais utilisé...  (EmbeddingsPerturbator et IdsPerturbator surcharge toutes les deux cette méthode donc je ne comprends pas vraiment l'intérêt d'avoir un comportement de base). 

De plus je suis ici les suggestions d'Antonin.

interpreto/attributions/base.py
HugoDeBosschere: Just to be clearer : the type of the tokenizer is now PreTrainedTokenizerBase in this current PR

interpreto/attributions/base.py
HugoDeBosschere: Comme c'est utile dans Insertion Deletion je l'ai remis.

tests/attributions/test_image_base.py
HugoDeBosschere: Yes je ferai ça :)

interpreto/attributions/perturbations/base.py
fanny-jourdan: Ok c'est bon pour moi alors :)

interpreto/attributions/perturbations/shap_perturbation.py
fanny-jourdan: Ok c'est clair merci ! @AntoninPoche je te pin pck tu avais fais celui de ShapTokenPerturbator dans mon souvenir donc peut etre que ça peut t'aider si tu veux le changer ?

interpreto/attributions/perturbations/sobol_perturbation.py
fanny-jourdan: Peut etre que justement tu peux te baser sur ce qu'ils ont fait dans Xplique pour Sobol sur les images :) On avait du beaucoup le modifier pour l'adapter au texte

tests/test_image_granularity.py
fanny-jourdan: Ok donc dans le doute laissons le jusqu'a ce qu'on ait la version finale

tests/visualizations/test_image_attribution.py
fanny-jourdan: Ok en vrai je pense que ça va si l'autre est nommé test_text_attribution (vu qu'ils sont dans le sous dossier vizu en plus)

tests/attributions/methods/test_attribution_methods_vision.py
fanny-jourdan: Oui avoir les explications dans les asserts c'est largement suffisant :)

tests/visualizations/test_text_attributions.py
```diff
@@ -1,3 +1,27 @@
+# MIT License
```
fanny-jourdan, line 1: Le code message automatique s'est mis une deuxieme fois dans ce fichier alors qu'il y était déjà

fanny-jourdan (approved):

PR #150, opened 2026-05-19
Simpler concepts splitter for classification and inputs to concepts attributions

AntoninPoche:
This PR can be reviewed commit by commit. About 5.5k line modifications are tutorial updates.

## Input-to-Concept Attributions

Adds the ability to attribute concept activations back to input tokens using perturbation-based methods. This connects the concept framework with the attribution framework. It is limited to the `*ForSequenceClassification` models.

The idea is that once the concept explainers have been fitted, we can extract their `inputs-to-concepts` and pass it to the attribution methods like a normal model.

`ModelWithSplitPoints` is overcomplex; I made a specific version that forces the split point and granularity to the [CLS] token.

### Key additions

- **`SplitSequenceClassification`**: A `ModelWithSplitPoints` specialization for classification models that auto-detects the classification head as a split point. Provides `inputs_to_activations` and `activations_to_outputs` for efficient concept pipelines.
- **`concept_explainer.inputs_to_concepts`**: Property on all concept explainers returning a `ModelForInputsToConcepts` bridge model (inputs → latent activations → concept activations). Quacks like a `PreTrainedModel` so attribution methods can consume it directly.
- **`InputsToConceptsInferenceWrapper`**: Inference wrapper handling the bridge model inside attribution explainers. Rejects gradient-based methods (incompatible with nnsight tracing).
- **`InputsToConceptsAttributionsExplainer`**: Task-specific attribution explainer auto-selected via `MultitaskExplainerMixin` when the model is a `ModelForInputsToConcepts`.

### Minimal example

```python
import torch
from interpreto import Occlusion, SplitSequenceClassification
from interpreto.concepts import SemiNMFConcepts

# 1. Wrap a classification model
split_model = SplitSequenceClassification(
    "nateraw/bert-base-uncased-emotion", batch_size=32, device_map="cuda"
)

# 2. Fit concepts
activations = split_model.get_activations(train_texts, tqdm_bar=True)
concept_explainer = SemiNMFConcepts(split_model, nb_concepts=20, device="cuda")
concept_explainer.fit(activations)

# 3. Attribute input tokens to concepts
explainer = Occlusion(concept_explainer.inputs_to_concepts, split_model.tokenizer)
results = explainer.explain("The market rallied on strong earnings.", targets=torch.arange(5))
```

### Also includes

- Tests for `SplitSequenceClassification` and input-to-concept attributions
- Documentation: new `interpretations/` docs section with concept attributions page, `SplitSequenceClassification` API page, updated overview

## Type of Change

[...]

PR #156, opened 2026-06-08
Concepts refacto & new generation specific splitter

AntoninPoche:
This PR should absolutely be reviewed commit by commit.

It touches many files due to renaming and some file moves. Also, it has a lot of lines because the notebooks are rerunning. But it is not that complex.

## Description

This PR was initiated by the observation that `ModelWithSplitPoint` was too complicated (with several split points, many granularities, and aggregation strategies). But in practice, either for classification or generation, we always use the same settings. It was initiated for classification in #150.

Hence, the goal of this PR is to remove dead code, simplify API, make computations more efficient

The main modifications are:
- Limit `ModelWithSplitPoints` to a single split point and remove the `split_point` parameter from any other object.
- `get_activations()` does not return a `dict[torch.Tensor]` anymore but a `tuple[torch.Tensor, torch.tensor | None]`. With a single split point, no need for a dictionary of activations. Only a tuple `(activations, predictions)` where the predictions can be `None`. This removes the need for the `get_split_activations()`
- Abstract splitting into a `BaseSplitter` and renaming every `model_with_split_points` attribute to `splitter`.
- Optimize how `nnsight` is used, also require `nnsight>=0.7.0,<0.8.0`
- Introduce a new `SplitterForGeneration` which forces the granularity to tokens, thus with no aggregations. Though there is still a parameter to include special tokens (but not padding).
- Adapt documentation to splitters' abstraction.
- Split the `interpreto.model_wrapping` module in `interpreto.attributions.inference_wrappers`, `interpreto.concepts.splitters` and move `llm_interface.py` to `interpreto.commons`. With new splitters and inference wrappers, the module was messy.
- Add `inputs_to_activations` to splitters.
- Rename `encode_activations` and `decode_concepts` to `activations_to_concepts` and `concepts_to_activations`.
- Add a `device` property and a `to()` method to concept explainers and link them directly to the `concept_model` device. The splitters are already user exposed, but users had no unified way to move the `concept_model` device. This also removes device management in interpretations.
- Adapt generation concept notebooks to modifications and rerun to validate validity.
- Fix the vendoring script.

## Breaking changes

- `ModelWithSplitPoints(split_points=[3])` is deprecated, not breaking. Use `ModelWithSplitPoints(split_point=3)` instead.
- `AnythingElse.split_point` or `AnythingElse(split_point=...)` will raise an error. The splitters are the only ones managing the split point.
- `get_activations()` output is not a dict but a tuple now.
- splitters do not have a `get_split_activations()` method anymore. It is not necessary; activations are the first element of the tuple.
```python
# Old and broken
activations_dict = splitter.get_activations(inputs)
activations = splitter.get_split_activationns

# New
activations, _  = splitter.get_activations(inputs)
```
- `from interpreto.model_wrapping import ModelWithSplitPoints` is not the correct path anymore. One should use `from interpreto import ModelWithSplitPoints` which was already the recommended way.
- Similarly, `from interpreto.model_wrapping.llm_interface import OpenAILLM` is not the correct path anymore. One should use `from interpreto.commons.llm_interface import OpenAILLM`.
- `encode_activations()` and `decode_concepts()` methods from concept explainers have been renamed to `activations_to_concepts()` and `concepts_to_activations()`. But these are mainly internal to the library.
- Interpretation methods do not have a `concept_model_device` anymore. But it should not change much, since the concept model is usually already on the device.
```python
# Old and broken
interpretation_method = TopKInputs(concept_explainer, concept_model_device="cuda")

# New
concept_explainer.to("cuda")  # only moves the `.concept_model`, not the `.splitter`
interpretation_method = TopKInputs(concept_explainer)
```

## Type of Change

## Checklist


interpreto/concepts/base.py
```diff
@@ -88,27 +87,26 @@ def __init__(
...
 
         self.concept_model = concept_explainer.concept_model
-        if self.concept_model.device != self.split_model.device:  # type: ignore  # TODO: add device to ConceptModelProtocol
-            self.concept_model.to(self.split_model.device)  # type: ignore
+        if self.concept_model.device != self.splitter.device:  # type: ignore  # TODO: add device to ConceptModelProtocol
+            self.concept_model.to(self.splitter.device)  # type: ignore
             self.concept_model.device = (  # type: ignore
-                self.split_model.device
+                self.splitter.device
             )  # Overcomplete models `.to()` method does not set the device
```
AntoninPoche: Adapt to new `concept_explainer.device`

interpreto/concepts/base.py
```diff
@@ -117,17 +115,17 @@ def eval(self):
...
             SimpleNamespace with a ``.logits`` attribute containing concept activations.
         """
-        activations: Float[torch.Tensor, "n d"] = self.split_model.inputs_to_activations(kwargs)
+        activations: Float[torch.Tensor, "n d"] = self.splitter.inputs_to_activations(kwargs)
 
-        # TODO: use `encode_activations` which should be renamed `activations_to_concepts`
+        # TODO: use `activations_to_concepts` which should be renamed `activations_to_concepts`
         concepts = self.concept_model.encode(activations)
         if isinstance(concepts, tuple):
             concepts = concepts[1]
```
AntoninPoche: Treat TODO

interpreto/concepts/base.py
```diff
@@ -149,7 +147,7 @@ def device(self, device: torch.device):
         Args:
             device (torch.device): wanted device (e.g., "cpu" or "cuda").
         """
-        self.split_model.to(device)  # type: ignore
+        self.splitter.to(device)  # type: ignore
         self.concept_model.to(device)  # type: ignore
```
AntoninPoche: Use `concept_explainer.device`.

interpreto/concepts/base.py
```diff
@@ -117,17 +115,17 @@ def eval(self):
 
     def resize_token_embeddings(self, new_num_tokens: int):
         """No-op: the concept model does not have token embeddings."""
-        self.split_model._model.resize_token_embeddings(new_num_tokens)
+        self.splitter._model.resize_token_embeddings(new_num_tokens)
 
     def __call__(self, **kwargs):
```
AntoninPoche, line 116: Maybe have an `concept_explainer.inputs_to_concepts` and just use it in the `__call__`

interpreto/attributions/inference_wrappers/__init__.py
AntoninPoche, line 1: The `interpreto/model_wrapping/` was split, so should `tests/model_wrapping/`

docs/api/concepts/splitters/model_with_split_points.md
```diff
@@ -2,6 +2,9 @@
 icon: material/code-json
 ---
 
+`ModelWithSplitPoints` now uses the singular `split_point` argument/property because only one split point is supported.
+The previous `split_points` argument/property remains temporarily available as a deprecated compatibility alias and emits a `DeprecationWarning` guiding users to `split_point`. It will be removed in version `0.6.0`.
```
fanny-jourdan: We're not sure if we should remove it from this version, are we? Especially if we move to version 1.0 once we've added multimodality.

interpreto/concepts/splitters/model_with_split_points.py
```diff
@@ -812,12 +812,12 @@ def get_activations(  # noqa: PLR0912  # ignore too many branches  # too many sp
...
         model_forward_kwargs: dict[str, Any] = {},
-    ) -> dict[str, LatentActivations] | dict[str, list[LatentActivations]]:
+    ) -> tuple[LatentActivations, torch.Tensor | None] | tuple[list[LatentActivations], list[torch.Tensor] | None]:
         """
 
-        Get intermediate activations for all model split points on the given `inputs`.
+        Get intermediate activations for the model split point on the given `inputs`.
 
-        Also include the model predictions in the returned activations dictionary.
+        Optionally include the model predictions in the returned tuple.
```
fanny-jourdan, line 708: I'm not sure I understand: in addition to the activations, do you output the predicted class? Or the logits? Is this just for classification?

interpreto/concepts/splitters/splitter_for_classification.py
```diff
@@ -259,10 +259,8 @@ def get_activations(  # type: ignore
             **kwargs: Unused, kept for API compatibility with ``ModelWithSplitPoints``.
 
         Returns:
-            dict[str, LatentActivations]: A dictionary with:
-                - ``self.classification_head_name``: Activations tensor of shape
-                  ``(n_samples, hidden_dim)``.
-                - ``"predictions"``: Predicted class indices of shape ``(n_samples,)``.
+            tuple[LatentActivations, torch.Tensor]: The activations tensor of shape
+                ``(n_samples, hidden_dim)`` and predicted class indices of shape ``(n_samples,)``.
```
fanny-jourdan, line 262: I thinked it can be None ?

docs/api/concepts/splitters/model_with_split_points.md
AntoninPoche: We can wait till version 1.0.0 if you prefer.

interpreto/concepts/splitters/model_with_split_points.py
AntoninPoche: It is classification-specific, and class logits are returned only if `include_predicted_classes` is True.

For `SplitterForClassification` it is always returned.

In `SplitterForGeneration` it is always None.

Would you prefer it to be only activations and eventually make a tuple if `include_predicted_classes` is True?

I would argue against so that we can always do `activations, _ = splitter.get_activations()`.

The other case would be that by default, predictions are never returned and only activations are returned:
```python
activations = splitter.get_activations(inputs)
activations, predictions = splitter.get_activations(inputs, include_predictions=True)
```

Which one do you prefer?

I think there are only these two possibilities because we want to be able to access the splitter's activations agnostically.

Note that we could extend the `include_predictions` to generation logits; it might be useful for VocabProj or Tokenchange. (But it needs to be done wisely; vocab size can be a real problem)

interpreto/concepts/splitters/splitter_for_classification.py
AntoninPoche: In `SplitterForClassification`, passing through the classification head costs little, so returning both does not significantly increase computation time.

But we could include an optional parameter to have it compute only when needed.

This is tightly linked with the previous comment.

interpreto/concepts/splitters/model_with_split_points.py
fanny-jourdan: Okay, thanks. I hadn't understood that the Boolean was set based on whether it was classification or generation (I knew you couldn't set it to True for generation, but I wasn't sure if it was fixed for classification or if we could change the Boolean ourselves). Maybe you mentioned it somewhere and I missed it, but if not, it would be good to state it clearly somewhere. I slightly prefer the approach where we just return the activations for the generation and the tuple for the classification, but I’m not opposed to keeping the current format (I’ll leave it up to you)

docs/api/concepts/overview.md
```diff
@@ -43,7 +43,7 @@ The API has five steps:
 
 `ModelWithSplitPoints` is a wrapper around your model to split it into different parts.
 It allows to obtain activations on a specific split point.
-The main method is thus `get_activations` which takes inputs and returns a dictionary of activations.
+The main method is thus `get_activations` which takes inputs and returns `(activations, predictions)`.
```
fanny-jourdan: It should be noted here that the prediction is only in the classification case (the doc seem general but if its only classification part you can ignore my comment)

interpreto/concepts/splitters/splitter_for_generation.py
```diff
@@ -0,0 +1,461 @@
...
+    Arguments:
+        model_or_repo_id (str | PreTrainedModel): A HuggingFace model ID or a
+            pre-loaded CausalLM instance.
+        split_point (str | int): The split location inside the model.
+        tokenizer (PreTrainedTokenizer | PreTrainedTokenizerFast | None): Tokenizer.
+            Required when providing a model instance.
+        config (PretrainedConfig | None): Model configuration.
+        batch_size (int): Batch size for batched operations.
+        device_map (torch.device | str | None): Device on which to load the model.
+        output_tuple_index (int | None): Index of the hidden state in a tuple output.
```
fanny-jourdan: Not sure to understand what is output_tuple_index

interpreto/concepts/splitters/splitter_for_generation.py
```diff
@@ -0,0 +1,461 @@
...
+        config: PretrainedConfig | None = None,
+        batch_size: int = 1,
+        device_map: torch.device | str | None = None,
+        output_tuple_index: int | None = None,
+        **kwargs,
+    ):
+        """Initialize a SplitterForGeneration model wrapper.
+
+        Raises:
+            TypeError: If ``model_or_repo_id`` is a PreTrainedModel that is not a CausalLM.
```
fanny-jourdan, line 117: model_or_repo_id can be also a LMHeadModel no ? (not only CausalLM)

fanny-jourdan: Ok it s done later

interpreto/concepts/splitters/splitter_for_generation.py
```diff
@@ -0,0 +1,461 @@
...
+        setattr(sp_module, module_out_name, layer_outputs)  # type: ignore
+
+    def _get_concept_output_gradients(
+        self,
+        inputs: list[str],
+        encode_activations: Callable[[LatentActivations], ConceptsActivations],
+        decode_concepts: Callable[[ConceptsActivations], LatentActivations],
+        targets: list[int] | None = None,
+        include_special_tokens: bool = False,
+        concepts_x_gradients: bool = False,
```
fanny-jourdan: concepts_x_gradients is by default False but we choose to have inputs_x_gradients as True by default. Why not here?

interpreto/concepts/splitters/splitter_for_generation.py
```diff
@@ -0,0 +1,461 @@
...
+                    tokens_mask,
+                )
+                del decoded_activations, raw_activations
+
+                # Get logits: (b, l, vocab) -> max over vocab -> (b, l) -> sum over samples -> (l,)
+                logits = self.output.logits.max(dim=-1)[0].sum(dim=0)
+
+                # Determine targets
+                if targets is None:
+                    current_targets = range(logits.shape[0])
```
fanny-jourdan, line 453: not sure to understand why

tests/model_wrapping/test_splitter_for_generation.py
```diff
@@ -0,0 +1,176 @@
...
+def test_loading_possibilities(gpt2_model, gpt2_tokenizer, bert_model, bert_tokenizer):
+    """Generation splitters can be loaded from repos or causal LM instances only."""
+    with pytest.raises(ValueError):
+        SFG(gpt2_model, split_point=SPLIT_POINT)
+
+    with pytest.raises(TypeError, match="not a causal language model"):
+        SFG(bert_model, split_point=SPLIT_POINT, tokenizer=bert_tokenizer)
+
+    split_model = SFG(gpt2_model, split_point=SPLIT_POINT, tokenizer=gpt2_tokenizer, device_map=DEVICE)
+    assert split_model.split_point is not None, "split_point should be resolved for a pre-loaded causal LM"
```
fanny-jourdan: Why are you just checking that it's not `None` instead of directly verifying that the form is correct? Especially since the error message doesn't seem to match exactly what you're testing here. (My comment applies to all the assert in the function.)

tests/concepts/splitters/test_splitter_for_generation.py
```diff
@@ -0,0 +1,176 @@
...
+        sentences,
+        include_all_tokens=include_all_tokens,
+    )
+    sample_wise_acts, sample_wise_predictions = split_gen.get_activations(
+        sentences,
+        include_all_tokens=include_all_tokens,
+        flatten_activations=False,
+    )
+
+    assert flattened_predictions is None
```
fanny-jourdan, line 121: maybe a message like "no prediction for generationc case" or something like that?

tests/concepts/splitters/test_splitter_for_generation.py
```diff
@@ -0,0 +1,176 @@
...
+        "Flattened activations do not match concatenated sample-wise activations"
+    )
+
+
+def test_get_activation_and_gradient(split_gen: SFG, sentences: list[str]):
+    """Activation and concept-output gradient shapes follow the generation splitter contract."""
+    hidden = split_gen._model.config.hidden_size
+    nb_concepts = 2 * hidden
+    initial = torch.randn(nb_concepts, hidden)
+    decoder_weights = torch.linalg.qr(initial)[0].to(DEVICE)
```
fanny-jourdan, line 145: what that means?

interpreto/concepts/splitters/splitter_for_generation.py
AntoninPoche: For some models, the output of layers is not a torch tensor but a tuple. The `manage_tuple_outputs()` method in the base splitter addresses this by extracting the first 3D tensor from the tuple as the raw activations.

The parameter corresponds to the index in this tuple. If not provided, it is detected following the above strategy.

Do you think we should leave it as a parameter? Or, no one will use it, and we would rather simplify the API?

interpreto/concepts/splitters/splitter_for_generation.py
AntoninPoche: That is indeed a miss. Even though this function should not be used by users. They should rather use `concept_explainer.concept_output_gradients()`.

docs/notebooks/generation_concept_tutorial.ipynb
```diff
@@ -548,17 +537,16 @@
...
    ],
    "source": [
-    "mwsp.to(\"cuda\")\n",
+    "splitter.to(\"cuda\")\n",
     "# (seq_len (out), seq_len (in), nb_concepts)\n",
     "local_importances = concept_explainer.concept_output_gradient(\n",
     "    inputs=sample,\n",
     "    targets=None,  # all predicted tokens\n",
-    "    activation_granularity=TOKEN,  # same throughout the notebook\n",
     "    concepts_x_gradients=False,  # based theoretically\n",
```
AntoninPoche: Should both be true I think

interpreto/concepts/splitters/splitter_for_generation.py
AntoninPoche: This means that if no targets are provided, we compute the contributions for all outputs, thus all output tokens.

It is used in the notebook, for example.

tests/model_wrapping/test_splitter_for_generation.py
AntoninPoche: Good point, I will update it.

tests/concepts/splitters/test_splitter_for_generation.py
AntoninPoche: This creates a placeholder encoder-decoder pair from matrix decomposition.

[`torch.linalg.qr`](https://docs.pytorch.org/docs/2.12/generated/torch.linalg.qr.html) performs `A = UW`. Where `A` is the initial random matrix.

docs/api/concepts/overview.md
```diff
@@ -39,36 +39,53 @@ concept_gradients = concept_explainer.concept_output_gradient(inputs=dataset)
...
+This step rely on the `.get_activations` method of the splitter.
+The idea is to compute a dataset of activations.
+To latter fit the concept model on.
+
+The function is quite simple for `SplitterForClassification` and `SplitterForGeneration`.
+However, for `ModelWithSplitPoints`, at this step, one needs to specify:
 
 - `activation_granularity`: specifies which of the `(n, l, d)` activations to return.
 It can be one of `CLS_TOKEN`, `ALL_TOKENS`, `TOKEN`, `WORD`, `SENTENCE`, or `SAMPLE`.
 Use `activation_granularity=ModelWithSplitPoints.activation_granularities.TOKEN` to specify it.
```
fanny-jourdan: Perhaps it would be clearer for users if the details specific to MWSP weren't included in the overview but only in the MWSP documentation.

interpreto/concepts/splitters/splitter_for_generation.py
fanny-jourdan: I don't think anyone will use it, so it's better if we detect it ourselves directly ;)

interpreto/concepts/splitters/splitter_for_generation.py
fanny-jourdan: I understood you want to compute the real target but i am suprised you use `range(logits.shape[0])` to do this

tests/concepts/splitters/test_splitter_for_generation.py
fanny-jourdan: Ok thanks !

interpreto/concepts/base.py
```diff
@@ -161,57 +159,52 @@ def to(self, device: torch.device, dtype: torch.dtype | None = None):
...
-        has_differentiable_concept_encoder (bool): Whether the `encode_activations` operation is differentiable.
+        has_differentiable_concept_encoder (bool): Whether the `activations_to_concepts` operation is differentiable.
     """
 
     has_differentiable_concept_encoder = False
 
     def __init__(
         self,
-        model_with_split_points: ModelWithSplitPoints,
+        model_with_split_points: BaseSplitter,
```
AntoninPoche: Forgot to rename parameter

interpreto/concepts/splitters/splitter_for_generation.py
AntoninPoche: The first dimension of the logits gives the sequence length. While the range converts it to [0, 1, 2, ..., seq_len].

interpreto/attributions/inference_wrappers/inference_wrapper.py
fanny-jourdan, line 1: Are inference wrappers only used in Attribution? (That doesn't surprise me; I'm just asking for confirmation, just in case.)

interpreto/model_wrapping/transformers_classes.py
```diff
@@ -1,88 +0,0 @@
-# MIT License
```
fanny-jourdan, line 1: Did you delete the entire file? If not, why was the DEEL header deleted?

interpreto/attributions/inference_wrappers/inference_wrapper.py
AntoninPoche: It is indeed the case.

interpreto/model_wrapping/transformers_classes.py
AntoninPoche: Yes, it was not used.

interpreto/concepts/splitters/base_splitter.py
```diff
@@ -285,6 +285,19 @@ def _manage_output_tuple(self, activations: torch.Tensor | tuple[torch.Tensor],
     # implementation at runtime, and get_latent_shape remains truly abstract
     # (its signature is uniform across all subclasses).
 
+    @abstractmethod
+    def inputs_to_activations(self, *args: Any, **kwargs: Any) -> Any:
+        """Un batched activation extraction from raw inputs.
```
fanny-jourdan: Unbatched instead of Un batched no?

interpreto/concepts/splitters/model_with_split_points.py
```diff
@@ -637,7 +701,7 @@ def get_activations(  # noqa: PLR0912  # ignore too many branches  # too many sp
         tqdm_bar: bool = False,
         include_predicted_classes: bool = False,
         flatten_activations: bool = True,
-        model_forward_kwargs: dict[str, Any] = {},
+        forward_kwargs: dict[str, Any] = {},
```
fanny-jourdan, line 702: Is there a reason why the name was changed?

interpreto/concepts/splitters/splitter_for_generation.py
```diff
@@ -422,11 +425,16 @@ def _get_concept_output_gradients(
                 activations: Float[torch.Tensor, "bg d"] = raw_activations.flatten(0, 1)[tokens_mask.flatten()]
 
                 # Encode activations into concepts
-                concept_activations: Float[torch.Tensor, "bg c"] = activations_to_concepts(activations)
+                concept_activations: Float[torch.Tensor, "bg c"] = activations_to_concepts(
+                    activations.to(dtype=torch.float32)
```
fanny-jourdan, line 427: why you forced float32?

tests/concepts/splitters/test_model_with_split_points.py
```diff
@@ -171,6 +171,7 @@ def test_get_activations_returns_tuple(splitted_encoder_ml: MWSP, sentences: lis
     """Activation extraction returns the split activations and optional predictions directly."""
     activations, predictions = splitted_encoder_ml.get_activations(sentences, activation_granularity=AG.ALL_TOKENS)
     assert isinstance(activations, torch.Tensor), "get_activations should return a tensor when flattened"
+    assert activations.dtype == torch.float32
```
fanny-jourdan, line 174: add a message like "activations have to be in float32" or something like this (here and for the other assert)

interpreto/concepts/splitters/model_with_split_points.py
```diff
@@ -858,6 +858,8 @@ def get_activations(  # noqa: PLR0912  # ignore too many branches  # too many sp
         activations: list[LatentActivations] = []
         predictions: list[torch.Tensor] = []
 
+        sp_module = self.get(self._split_point)
```
fanny-jourdan, line 854: what is it?

interpreto/_vendor/overcomplete/_sync_overcomplete.py
fanny-jourdan, line 1: I'm not sure I understand the changes here, but I trust you

docs/notebooks/generation_concept_tutorial.ipynb
fanny-jourdan, line 1: I would have put the Probe notebooks in notebooks.examples rather than directly in notebooks to avoid overloading it.

docs/notebooks/classification_concept_tutorial.ipynb
fanny-jourdan, line 1: Not for this directly but maybe it can be cool to rename attribution_walkthrough.ipynb in attribution_tutorial.ipynb to fit better with the rest ?

docs/notebooks/classification_concept_tutorial.ipynb
```diff
@@ -63,7 +63,7 @@
    "source": [
     "from interpreto import SplitSequenceClassification\n",
```
fanny-jourdan: I have this error when I run: 
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[9], [line 1](vscode-notebook-cell:?execution_count=9&line=1)
----> [1](vscode-notebook-cell:?execution_count=9&line=1) from interpreto import SplitSequenceClassification
      3 splitter = SplitSequenceClassification(
      4     model_or_repo_id="textattack/distilbert-base-uncased-ag-news",
      5     device_map="cuda",
      6     batch_size=64,
      7 )

ImportError: cannot import name 'SplitSequenceClassification' from 'interpreto' (/home/fanny.jourdan/dev/interpreto/docs/notebooks/../../interpreto/__init__.py)
```

docs/notebooks/generation_concept_tutorial.ipynb
fanny-jourdan, line 1: this cell : 
```
splitter.to("cuda")
#(seq_len (out), seq_len (in), nb_concepts)
local_importances = concept_explainer.concept_output_gradient(
    inputs=sample,
    targets=None,  # all predicted tokens
    concepts_x_gradients=False,  # based theoretically
    normalization=False,  # for each output token, the sum of the absolute values of the importance is equal to 1
    batch_size=64,
)[0]  #only one sample
splitter.to("cpu")

#we take the sum over the input sequence dimension, has we focus on the concept-output relationship
#(seq_len (out), nb_concepts)
local_importances = local_importances.abs().sum(dim=1)
print(f"{local_importances.shape = }")
```
looks a little messy

fanny-jourdan, line 1: In the cell `5.3 Visualization`, we have this error :
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[11], [line 11](vscode-notebook-cell:?execution_count=11&line=11)
      8 to_interpret = [c.item() for c in to_interpret if c.item() not in interpretations.keys()]
     10 # interpret these concepts
---> [11](vscode-notebook-cell:?execution_count=11&line=11) new_interpretations = interpretation_method.interpret(
     12     inputs=imdb,
     13     latent_activations=activations,
     14     concepts_indices=to_interpret,
     15 )
     16 interpretations.update(new_interpretations)

File ~/dev/interpreto/docs/notebooks/../../interpreto/concepts/interpretations/topk_inputs.py:276, in TopKInputs.interpret(self, concepts_indices, inputs, latent_activations, concepts_activations)
    239 def interpret(
    240     self,
    241     concepts_indices: int | list[int] | Literal["all"] = "all",
   (...)    244     concepts_activations: ConceptsActivations | None = None,
    245 ) -> Mapping[int, Any]:
    246     """
    247     Give the interpretation of the concepts dimensions in the latent space into a human-readable format.
    248     The interpretation is a mapping between the concepts indices and a list of inputs allowing to interpret them.
   (...)    273 
    274     """
    275     sure_concepts_indices, granular_inputs, sure_concepts_activations, _ = (
--> [276](https://vscode-remote+ssh-002dremote-002bdeel17.vscode-resource.vscode-cdn.net/home/fanny.jourdan/dev/interpreto/docs/notebooks/~/dev/interpreto/interpreto/concepts/interpretations/topk_inputs.py:276)         self.get_granular_inputs_and_concept_activations(
...
    157         f"At least one concept index out of bounds. `max(concepts_indices)`: {max(concepts_indices)} >= {concepts_activations.shape[1]}."
    158     )
    160 return concepts_indices

ValueError: max() iterable argument is empty
```

fanny-jourdan, line 1: There are also a part called `7. Using your own LLM interface` but there are no example and only a function.

docs/notebooks/examples/classification_probe_tutorial.ipynb
fanny-jourdan, line 1: Same importation error than classification_concept_tutorial.ipynb

fanny-jourdan (approved):
Great job, Antonin! The changes were necessary and well done ;) 
I've added a few comments to the code (and some questions to make sure I understood it correctly), and most importantly, some things to fix in the notebook that aren't running properly. I'll leave it to you to take care of these small issues, and then you can merge!

docs/notebooks/examples/generation_demonstration.ipynb
fanny-jourdan, line 1: there are error : `ModuleNotFoundError: No module named 'openai'` please add a filter like the concept_generation_tutorial

docs/notebooks/examples/classification_demonstration.ipynb
fanny-jourdan, line 1: always the same error: `AttributeError: module 'interpreto' has no attribute 'SplitSequenceClassification'`

interpreto/concepts/splitters/model_with_split_points.py
AntoninPoche: Harmonization with other places in the repo.

interpreto/concepts/splitters/splitter_for_generation.py
AntoninPoche: The overcomplete concept models are in float32, so it needs to be forced...

interpreto/concepts/splitters/model_with_split_points.py
AntoninPoche: All splitters inherit from `nnsight.LanguageModel` in which modules correspond to the initial model.

Here, sp_module corresponds to the module where we split and extract the activations. This operation was done each time in the loop, so I moved it out of the loop.

It is on this that we do `activations = sp_module.output.save()`. (Well, there are weird nns_output and uple outputs, but the idea is there.)

interpreto/_vendor/overcomplete/_sync_overcomplete.py
AntoninPoche: In summary, the problem was that each week a PR was opened because the date the script was run (in the workflow) was different from the initial date the code was vendored.

To solve this, I added an early exit to the script. Basically, if the last overcomplete matches the vendored commit, we do nothing.

docs/notebooks/classification_concept_tutorial.ipynb
AntoninPoche: It is weird that I missed this. I renamed it `SplitterForClassification` now. I will fix the notebook.

docs/notebooks/generation_concept_tutorial.ipynb
AntoninPoche: I think the comments are necessary (while there a few mistakes that I will correct).

I can skip the default parameter, though.

AntoninPoche: I will rerun it and see if I have the same mistake.

AntoninPoche: Yes, because it is basically `LLMLabel(..., llm_interface=MyLLMInterface, ...)` and exactly the same as the rest then.

Do you think it is necessary to include?

AntoninPoche: I did not get this error. Did you reach other errors earlier in the notebook?

fanny-jourdan: Yeah, I think it would be clearer to include the running code anyway :)

fanny-jourdan:
I reran all the notebooks with your changes, and everything is working fine now. :)

